PR #2, opened 2023-04-04
Add output parsers, prompt templates, memory, llmChain and example

FluffyKebab:
- Add an interface for output parsers
- Implement a structured parser that implements this interface
- Add prompt templates with f-strings
- Add a memory interface analogous to the abstract class BaseMemory in the JavaScript implementation
- Implement a chain interface and a function for calling chains
- Add the LLM chain as a chain
- Add tests for the changes made


ChristianSch:
@tmc how do you feel about merging this? :)

PR #4, opened 2023-04-17
Move namespace in pinecone index from a struct field to function argument

FluffyKebab:
Move the namespace field in the pinecone index from a struct field to function argument. This will give users the option to have one pinecone index for multiple namespaces, reducing the api calls done when creating a new pinecone client.


tmc (approved):
I’ll have some suggestions to improve the api of this package but this seems like a useful change in the meantime.

PR #8, opened 2023-04-18
Feature/adding agents

Abraxas-365:


agents/baseAgent.go
```diff
@@ -0,0 +1,63 @@
+package agents
```
tmc, line 1: every new package should have a doc.go file (I need to write up contribution guidelines, I will soon).

agents/baseAgent.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"strings"
+
+	"github.com/tmc/langchaingo/chains"
+	"github.com/tmc/langchaingo/llms"
+	"github.com/tmc/langchaingo/tools"
+)
+
+type Agent struct {
```
tmc, line 11: in Go we try to avoid "stuttering" -> I'd call this BaseAgent before "Agent" and the filename should reflect that -> base_agent.go. These exported types should all have go-idiomatic comments.

agents/baseAgent.go
```diff
@@ -0,0 +1,63 @@
...
+	Llm   llms.LLM
+	Chain chains.Chain
+	Tools []tools.Tool
+}
+
+type AgentInterface interface {
+	Run(input string) string
+}
+
+func FromLlmAndTools(llm llms.LLM, chain chains.Chain, tools []tools.Tool) Agent {
```
tmc, line 21: constructors in go almost always are in the form of (package).NewXXX where XXX is the type in question.

agents/baseAgent.go
```diff
@@ -0,0 +1,63 @@
...
+	return Agent{
+		llm,
+		chain,
+		tools,
+	}
+}
+
+func GetActionAndInput(input string) (action, actionInput string) {
```
tmc, line 29: All exported symbols should have comments.

Separately, this doesn't feel like it should live directly in this package namespace, further, generally, we avoid named returns in Go.

agents/baseAgent.go
```diff
@@ -0,0 +1,63 @@
...
+		if strings.HasPrefix(field, "Action: ") {
+			action = strings.TrimPrefix(field, "Action: ")
+		} else if strings.HasPrefix(field, "Action Input: ") {
+			actionInput = strings.TrimPrefix(field, "Action Input: ")
+		}
+	}
+	return action, actionInput
+}
+
+func GetObservation(action string, actionInput string, tools []tools.Tool) string {
```
tmc, line 41: All exported symbols should have comments.

agents/baseAgent.go
```diff
@@ -0,0 +1,63 @@
...
+		if tool.Name() == strings.Trim(action, " ") {
+			observation := "\nObservation: " + tool.Run(actionInput) + "\n"
+			return observation
+
+		}
+	}
+	return ""
+}
+
+func GetFinalAnswer(text string) string {
```
tmc, line 52: as above, should live in some sort of helper package and should have a comment

tools/googlesearch/searchTool.go
```diff
@@ -0,0 +1,48 @@
+package googlesearch
```
tmc, line 1: new packages should have doc.go files, please keep all filename lowercase, use underscores if you'd like.

tools/googlesearch/searchTool.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"strings"
+
+	"github.com/tmc/langchaingo/tools"
+	"github.com/tmc/langchaingo/utils/serpapi"
+)
+
+type SearchTool struct {
```
tmc, line 10: All exported symbols should have comments.

tools/googlesearch/searchTool.go
```diff
@@ -0,0 +1,48 @@
...
+	name        string
```
tmc, line 11: it's odd to export a type but then make the fields unexported, can you explain your rationale?

tools/googlesearch/searchTool.go
```diff
@@ -0,0 +1,48 @@
...
+	description string
+	serpapi     *serpapi.SerpapiWrapper
+}
+
+func NewSearchTool() (tools.Tool, error) {
```
tmc, line 16: All exported symbols should have comments. You should prefer to return a concrete type over an interface (as it will/can satisfy that interface). Please see https://bryanftan.medium.com/accept-interfaces-return-structs-in-go-d4cab29a301b

tools/googlesearch/searchTool.go
```diff
@@ -0,0 +1,48 @@
...
+        "Useful for when you need to answer questions about current events. "
+		"Always one of the first options when you need to find information on internet"
+        "Input should be a search query."
+		"Priority: 1"`,
+		serpapi: serpapi,
+	}, nil
+
+}
+
+func (s *SearchTool) Name() string {
```
tmc, line 33: there's no reason in Go to wrap a field like this, in general. Instead you should just export the field and avoid writing a method in this manner.

tmc (commented):
This is a partial review but hopefully you can extrapolate out.

I really appreciate the contribution but also am trying to drive a consistent approach to api design in this codebase. Thanks so much and I'm looking to iterate on the review!

PR #9, opened 2023-04-18
Feature/adding stop words

Abraxas-365:
adding the suport to put stop words, si the openIa api could stop , this si going to be usefull for the agents

### DIFF
--- a/agen/baseAgent.go
+++ b/agen/baseAgent.go
@@ -1,0 +1,66 @@
+package agent
+
+import (
+	"strings"
+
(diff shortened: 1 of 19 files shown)


tmc:
Thanks for your contribution! this is adding a bit more than I'd expect at a time -> the agent changes alone should be their own proposal. Could you break this down into smaller parts to simplify review? 🙏

Abraxas-365:
Sorry Im goint to create a new PR with just the suport for stopWords

PR #10, opened 2023-04-18
examples: adding support for .env in examples

mfreeman451:


.gitignore
```diff
@@ -0,0 +1,4 @@
+.env
+*.sum
```
tmc, line 2: I don't think .sum should be an ignored pattern

.gitignore
```diff
@@ -0,0 +1,4 @@
...
+vendor
```
tmc, line 3: could you place this and the .idea into a local gitignore rule as to keep the shared repo-level .gitignore as simple as possible?

embeddings/openai_test.go
```diff
@@ -9,20 +9,20 @@ import (
 func TestOpenaiEmbeddings(t *testing.T) {
 	e, err := embeddings.NewOpenAI()
 	if err != nil {
-		t.Errorf("Unexpected error creating openai embeddings struct: %e", err)
+		t.Errorf("Unexpected error creating openai embeddingsVar struct: %e", err)
```
tmc, line 12: This should simply be `t.Error(err)`

embeddings/openai_test.go
```diff
@@ -9,20 +9,20 @@ import (
...
 	}
 
 	_, err = e.EmbedQuery("Hello world!")
 	if err != nil {
 		t.Errorf("Unexpected error embed query: %e", err)
 	}
 
-	embeddings, err := e.EmbedDocuments([]string{"Hello world", "The world is ending", "bye bye"})
+	embeddingsVar, err := e.EmbedDocuments([]string{"Hello world", "The world is ending", "bye bye"})
```
tmc, line 20: I'd just call this var "e" instead.

examples/QA-example/main.go
```diff
@@ -21,6 +22,12 @@ var dimensions = 1536
 var numDocsInReq = 5
 
 func main() {
+	// load .env with joho/godotenv
+	err := godotenv.Load()
+	if err != nil {
+		log.Println("Error loading .env file")
+	}
```
tmc, line 29:
```suggestion
	if err := godotenv.Load(); err != nil {
		log.Println("issue .env file:", err)
	}
```

can you reflect these in the other examples? How does this end up if .env doesn't exist?

go.work
```diff
@@ -1,3 +1,7 @@
 go 1.20
```
tmc, line 1: note: I omitted (and will continue to omit) this file from source control but I'm open to some tooling to help out folks w pointing go.work files at the main tree.

tmc (commented):
Thanks for your contribution! I left some feedback relevant to this change.

tmc:
@mfreeman451 can you get this rebased/de-conflicted? don't want it to languish.

tmc:
Closing this given the age and conflicts.

PR #14, opened 2023-04-18
exp: Move recent contributions into exp subdir for now

tmc:
I merged a bit hastily in PR 2 and there are a number of contribution expectations I'd like to set and in service of doing that I'm going to move current packages under an "exp" dir (for experimental) with the intention of moving them out of that directory as they meet some contribution standards.


FluffyKebab:
Can you explain what needs to be changed in the packages?

PR #15, opened 2023-04-18
Feature/stop words

Abraxas-365:
Sorry fro previous PRs this is just the stop words support


FluffyKebab:
In the typescript version stop is just a key in the input values to a chain. I feel like that is cleaner than adding a new argument to the call method of chains.

Abraxas-365:
Great, I can fix that

Issue #11, opened 2023-04-18
Improve CONTRIBUTING.md


tmc:
As interest is spooling up here it's more important to have clear contribution guidelines.

In scope should include:
* Pull request sizing and scoping.
* PR title and description norms.
* Code quality expectations
* Testing expectations

tmc:
This has been addressed.

PR #16, opened 2023-04-18
docs: added CONTRIBUTING.md and CODE_OF_CONDUCT.md

nekomeowww:
## Summary

Done #11 

## Details

I reached out to find some good examples of `CONTRIBUTING.md` and find out the [CONTRIBUTING.md of Github Docs](https://github.com/github/docs/blob/main/CONTRIBUTING.md) and [Contributing.md Generator](https://generator.contributing.md/) seems fit the intial needs. So I combined these two and tweaked the structure of it.

But I still think it is important to find a way to state out what kind of the Golang coding guidelines we may follow, such as Uber's Golang Guideline or so. It might be optional to @tmc to decide whether we should add it or not.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,69 @@
+# Code of Conduct - langchaingo
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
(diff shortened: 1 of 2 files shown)


nekomeowww:
Me too!

PR #17, opened 2023-04-18
fix: rename test exp packages and fix all test compiling error

nekomeowww:
## Summary

Part of the fixes of #18.

## Details

- Placed the test files near by the code file it tested instead of nesting with one more `test` sub-dir in order to meet the idiomatic requirements
- Removed suffix `_test` from the package names of test files in order to align the package names and meet the idiomatic requirements
- Fixed all the imports issue by adding `exp` sub-dir into the import path


tmc:
@nekomeowww thanks for this fix! I see the main benefit of aligning test package names with the tested package name as a means of doing greybox style testing of unexported functions/types/fields. In this/our case I can't help but think we should preserve "_test" in the package names (but very much agree with moving the test files into the corresponding directories).

nekomeowww:
> @nekomeowww thanks for this fix! I see the main benefit of aligning test package names with the tested package name as a means of doing greybox style testing of unexported functions/types/fields. In this/our case I can't help but think we should preserve "_test" in the package names (but very much agree with moving the test files into the corresponding directories).

Yes it is for unexported functions/types/fields since we are utilities library and we will only export the needed ones to users while we may have a lot of helper functions to implement the algorithms.

nekomeowww:
Btw, thanks for helping on fixing those incorrect tests. I didn't include it because I thought it would be nice to separate them into different PRs to speed up the review 😆

PR #20, opened 2023-04-18
Feature/stop words

Abraxas-365:
Pls check if you like this implementation

### DIFF
--- a/exp/chains/llmChain.go
+++ b/exp/chains/llmChain.go
@@ -13,15 +13,17 @@ type LLMChain struct {
 	OutputKey    string
 	Memory       memory.Memory
 	OutputParser outputParsers.OutputParser
+	StopWords    []string
 }
 
-func NewLLMChain(llm llms.LLM, prompt prompts.Template) LLMChain {
+func NewLLMChain(llm llms.LLM, prompt prompts.Template, stopWords []string) LLMChain {
 	chain := LLMChain{
 		prompt:       prompt,
 		llm:          llm,
 		OutputKey:    "text",
 		OutputParser: outputParsers.NewEmptyOutputParser(),
 		Memory:       memory.NewEmptyMemory(),
+		StopWords:    stopWords,
 	}
 
 	return chain
--- a/exp/chains/llmChain.go
+++ b/exp/chains/llmChain.go
@@ -34,14 +34,13 @@ func (c LLMChain) GetMemory() memory.Memory {
 }
 
 func (c LLMChain) Call(values ChainValues) (ChainValues, error) {
-	//TODO: stop
 
 	promptValue, err := c.prompt.FormatPromptValue(values)
 	if err != nil {
 		return ChainValues{}, err
 	}
 
-	generations, err := c.llm.Generate([]string{promptValue.String()})
+	generations, err := c.llm.Generate([]string{promptValue.String()}, c.StopWords)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


FluffyKebab:
I don't think stop words should be a field in the LLM chain. Rather it should be a part of the chain values given as an input when calling the chain.

Check out how it is done in the typescript implementation:
async _call(values: ChainValues): Promise<ChainValues> {
    let stop;
    if ("stop" in values && Array.isArray(values.stop)) {
      stop = values.stop;
    }
    const promptValue = await this.prompt.formatPromptValue(values);
    const { generations } = await this.llm.generatePrompt([promptValue], stop);
    return {
      [this.outputKey]: await this._getFinalOutput(generations[0], promptValue),
    };
  }
So we should check if the chainValues have a stop key and then use those stop words.

Abraxas-365:
Ohh I get it, but, but those ChainValues are formatted here 
```go
func (c LLMChain) Call(values ChainValues) (ChainValues, error) {

	promptValue, err := c.prompt.FormatPromptValue(values)
	if err != nil {
		return ChainValues{}, err
	}

	generations, err := c.llm.Generate([]string{promptValue.String()}, c.StopWords)
	if err != nil {
		return ChainValues{}, err
	}

	finalOutput, err := c.OutputParser.ParseWithPrompt(generations[0].Text, promptValue)
	if err != nil {
		return ChainValues{}, err
	}

	return ChainValues{c.OutputKey: finalOutput}, nil
}
```
the **c.prompt.FormatPromptValue(values)** format it as parto of the prompt template 
```go
type PromptTemplate struct {
	partialVariables map[string]any
	inputVariables   []string
	template         string
	templateFormat   string
	validateTemplate bool
}
```
I think its a little bit weird , or maybe there is something im not seeing sorry

Abraxas-365:
also in the ts they are sending the stop word to the here 
```ts
 const { generations } = await this.llm.generatePrompt([promptValue], stop);
```
in my approach im setting the stopWord as part of the chain , and both ts and go are sending the stop in the **llms.Generate**

FluffyKebab:
I am thinking somthing like this

```go
func (c LLMChain) Call(values ChainValues) (ChainValues, error) {

	promptValue, err := c.prompt.FormatPromptValue(values)
	if err != nil {
		return ChainValues{}, err
	}

    stopWords, ok := values["stop"]

    // then check that stopWords exists and is a string slice 

	generations, err := c.llm.Generate([]string{promptValue.String()}, /* use stop words gotten here */)
	if err != nil {
		return ChainValues{}, err
	}

	finalOutput, err := c.OutputParser.ParseWithPrompt(generations[0].Text, promptValue)
	if err != nil {
		return ChainValues{}, err
	}

	return ChainValues{c.OutputKey: finalOutput}, nil
}
```

This will make creating llmChains simpler and is closer to the typescript implementation.

Abraxas-365:
Okey, Im not 100% convinced though, bc this method is a little less verbose , also bc we are mixing the same mechanism "ChainValues"  for stop and promp creation, also ChainValues is a map , maybe if it was a struct?  I fell the other way a easy to create but harder to know wahts happening and the params , whould like it a little more verbose

Abraxas-365:
But both approaches are fine, if you prefere the other way I'll doit :)

Abraxas-365:
I made the changes to work as the Ts one, should I solve the confilcts?

PR #13, opened 2023-04-18
feat: adding golint and a Makefile

mfreeman451:
1. Added a `Makefile` that includes tests for golint, ineffassign, gosec, gocov, etc.
2. Adding `coverage.yml` to ci for gocov
3. Added `golintci-lint.yml` to ci for golint
4. Updated a test so it would pass gosec check

PR #22, opened 2023-04-19
Add support for local LLMs

rdbell:
This implements support for local LLMs.

Tested with various local completion/chat engines.

```
╰─ 〉LOCAL_LLM_BIN="/path/to/alpaca/chat" LOCAL_LLM_ARGS="-m /path/to/alpaca/ggml-alpaca-7b-q4.bin -p" go run .
A square has four equal sides, so it has 4 sides in total.
```

You can test with any local binary if you don't have a local LLM set up on your machine.

For example, `LOCAL_LLM_BIN="/usr/bin/say" LOCAL_LLM_ARGS="-v 'Good News'" go run main.go` works on macOS.

### DIFF
--- a/llms/local/localllm.go
+++ b/llms/local/localllm.go
@@ -1,0 +1,61 @@
+package local
+
+import (
+	"context"
+	"errors"
+	"os"
+
+	"github.com/tmc/langchaingo/llms"
+	"github.com/tmc/langchaingo/llms/local/internal/localclient"
+)
+
+const localLLMBinVarName = "LOCAL_LLM_BIN"
+const localLLMArgsVarName = "LOCAL_LLM_ARGS"
+
+var ErrEmptyResponse = errors.New("no response")
+var ErrMissingBin = errors.New("missing the local LLM binary path, set the LOCAL_LLM_BIN environment variable")
+
+type LLM struct {
+	client *localclient.Client
+}
+
+var _ llms.LLM = (*LLM)(nil)
+
+func (o *LLM) Call(prompt string) (string, error) {
+	r, err := o.Generate([]string{prompt})
+	if err != nil {
+		return "", err
+	}
+	if len(r) == 0 {
+		return "", ErrEmptyResponse
+	}
+	return r[0].Text, nil
+}
+
+func (o *LLM) Generate(prompts []string) ([]*llms.Generation, error) {
+	result, err := o.client.CreateCompletion(context.TODO(), &localclient.CompletionRequest{
+		Prompt: prompts[0],
+	})
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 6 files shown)


llms/local/localllm.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"context"
+	"errors"
+	"os"
+
+	"github.com/tmc/langchaingo/llms"
+	"github.com/tmc/langchaingo/llms/local/internal/localclient"
+)
+
+const localLLMBinVarName = "LOCAL_LLM_BIN"
```
tmc, line 12: nit: this and the following var blocks can be like:
```
const (
    ...
```

llms/local/localllm.go
```diff
@@ -0,0 +1,61 @@
...
+var ErrEmptyResponse = errors.New("no response")
+var ErrMissingBin = errors.New("missing the local LLM binary path, set the LOCAL_LLM_BIN environment variable")
+
+type LLM struct {
+	client *localclient.Client
+}
+
+var _ llms.LLM = (*LLM)(nil)
+
+func (o *LLM) Call(prompt string) (string, error) {
```
tmc, line 24: can you add comments to these methods?

llms/local/localllm.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/tmc/langchaingo/llms/local/internal/localclient"
+)
+
+const localLLMBinVarName = "LOCAL_LLM_BIN"
+const localLLMArgsVarName = "LOCAL_LLM_ARGS"
+
+var ErrEmptyResponse = errors.New("no response")
+var ErrMissingBin = errors.New("missing the local LLM binary path, set the LOCAL_LLM_BIN environment variable")
+
+type LLM struct {
```
tmc, line 18: merits a comment

llms/local/localllm.go
```diff
@@ -0,0 +1,61 @@
...
+		return nil, err
+	}
+	return []*llms.Generation{
+		{Text: result.Text},
+	}, nil
+}
+
+func New() (*LLM, error) {
+	// Require the user to set the path to the local LLM binary
+	binPath := os.Getenv(localLLMBinVarName)
```
tmc, line 49: we can layer this on but instead of requiring a specific path we should probably use [exec.LookPath](https://pkg.go.dev/os/exec#LookPath)

tmc (commented):
Awesome contribution!

This looks really great just have a few small suggestions.

tmc:
I'll address my comments in a follow-up.

rdbell:
Thanks. I'll submit another PR with the suggestions mentioned.

The `golangci-lint` pipeline PR #13 should be useful for enforcing stuff like this.

PR #25, opened 2023-04-19
Expand tilde in path

rdbell:
- Add expansion of `~` to home directory in path lookups.
- Add expansion of environment variables in path lookups. Not typically useful if executed from common shells, as expansion is normally done during command execution, but may handle edge cases like if the binary is called by some other binary.

The following formats should now be working:

```
/Users/xxx/path/to/binary_name # Full path
$BASE/path/to/binary_name      # Env variable expansion
~/path/to/binary_name          # Tilde expansion
./path/to/binary_name          # Relative path
binary_name                    # $PATH lookup
```

This should handle most typical path inputs but there may be some go package that does this more thoroughly.

Additional considerations may be useful for Windows environments.

Also fixed a comment in `openaiclient.go`


llms/local/localllm.go
```diff
@@ -60,15 +60,27 @@ func (o *LLM) Generate(prompts []string) ([]*llms.Generation, error) {
...
-		return nil, errors.Join(ErrMissingBin, err)
-	}
+	binPath := os.Getenv(localLLMBinVarName)
 
 	// Allow the user to pass CLI arguments to the local LLM binary (optional)
 	args := os.Getenv(localLLMArgsVarName)
-	c, err := localclient.New(path, args)
+
+	// Expand tilde in the path
+	if binPath[0] == '~' {
```
nekomeowww, line 69:
```suggestion
	if strings.HasPrefix(binPath, "~") {
```

llms/local/localllm.go
```diff
@@ -60,15 +60,27 @@ func (o *LLM) Generate(prompts []string) ([]*llms.Generation, error) {
...
+		binPath = "$HOME" + binPath[1:]
```
nekomeowww, line 70: Can you write some tests against it?
And it should be proper to use `filepath.Join(...)` to concat the path elements.

```suggestion
		binPath = filepath.Join("$HOME", strings.TrimPrefix(binPath, "~"))
```

Btw, if you choose to accept my suggestion, you should also import the `filepath` package as well within the `import` section.

llms/local/localllm.go
tmc: can we put this into a helper function?

nekomeowww: Let me do some research on filepath handling... there might be existing libraries can help us.

nekomeowww: Hmm, it looks like we have to wrap the helper functions by ourselves.

@rdbell How about creating a new `util` package and place the `~` determine and resolving function? And don't forget to create a relevant `doc.go` file with the package comments in it.

@tmc What do you think?

rdbell:
- Moved path finding to a separate helper function in the `util` package
- Switched to to `strings.HasPrefix` instead of `path[0] ==`
- Switched to `os.UserHomeDir()` instead of relying on `$HOME`
- Added tests

Ready on my end.

rdbell:
Re-introduced parameter parsing for local LLMs. Without this `exec.Command` will fail when attempting to pass in multiple parameters.

rdbell:
I'm not sure if there's a good way to satisfy the linter's `cyclop` warning in this case.

Something like this would work:

```
func isSplitChar(char rune, inDoubleQuote, inSingleQuote bool) bool {
    return char == ' ' && !inDoubleQuote && !inSingleQuote
}

func processQuote(char, quoteChar rune, inQuote bool) bool {
    return char == quoteChar && !inQuote
}

func stringSplit(args string) []string {
    var parts []string
    var currentPart string
    var inDoubleQuote, inSingleQuote bool

    for _, char := range args {
        switch {
        case isSplitChar(char, inDoubleQuote, inSingleQuote):
            if currentPart != "" {
                parts = append(parts, currentPart)
                currentPart = ""
            }
        case processQuote(char, '"', inSingleQuote):
            inDoubleQuote = !inDoubleQuote
        case processQuote(char, '\'', inDoubleQuote):
            inSingleQuote = !inSingleQuote
        default:
            currentPart += string(char)
        }
    }

    if currentPart != "" {
        parts = append(parts, currentPart)
    }

    return parts
}
```

But I'm not sure that's actually helpful for increasing readability.

rdbell:
Resolving this with `nolint: cyclop` for now.

I prefer to avoid using `nolint` whenever possible, but I'm not sure that it's appropriate to make changes that would inhibit readability and add single-use functions here.

Open to suggestions, otherwise ready to merge on my end.

tmc:
@rdbell can you rebase this?

Issue #18, opened 2023-04-18
Need to fix and improve the implementations and tests under `exp` sub-directory


nekomeowww:
## Summary

It is impossible to develop, fix, and improve now because of the #14 changes didn't rename the imported packages in tests of `exp` packages, and the test code and file structure isn't idiomatic for Golang.

And also, I found out there are a lot of bugs and critical logical issues of implementation and tests within the packages under the exp sub-dir, however, I have to go step by step and break down things apart in order to boost up the review speed.

## TODOs

- [x] Fix structuring issues (#17)
- [x] Fix import issues (#17)
- [x] Fix testing logical issue (such as t.FailNow() not called if error occurred and will interfere the further steps, didn't assert whether the if statement branch is entered or not, etc.) (#28 )
- [x] Fix the lint issues (#27, #28)
- [x] Fix the implementation issues (#27)

tmc:
@nekomeowww can we consider this closed now? we can talk out testify separately

PR #27, opened 2023-04-19
chore: improve code qualities and use idiomatic way to implementation

nekomeowww:
## Description

Improvements of #18 

- Removed unnecessary code
- Use `http.StatusXxx` instead of hardcoded numbers when checking http status
- Replaced incorrect use of `fmt.Printf(..)` with `log.Printf(...)` with idiomatic prefixes `[WARN]`
- Replaced incorrect use and deprecated `ioutil.ReadAll(...)` with `*bytes.Buffer` and `io.Copy(...)` for better performance


tmc:
LGTM

PR #30, opened 2023-04-19
chains: add support for stop words

Abraxas-365:
Implementing stop words as the Ts version


tmc:
Can you link to the ts/python inspiration here and include test cases?

Abraxas-365:
this is the ts version
```ts
async _call(values: ChainValues): Promise<ChainValues> {
    let stop;
    if ("stop" in values && Array.isArray(values.stop)) {
      stop = values.stop;
    }
    const promptValue = await this.prompt.formatPromptValue(values);
    const { generations } = await this.llm.generatePrompt([promptValue], stop);
    return {
      [this.outputKey]: await this._getFinalOutput(generations[0], promptValue),
    };
  }
```

Abraxas-365:
Where should i put the test cases?

tmc:
Right alongside the code in _test.go files, reference other existing (non-exp) code.

Abraxas-365:
@tmc Hi i hace to options of doing the stopWords implementation which one you prefer? i would prefer as part of the function Param , but what is your opinion?
# The first one is putting the stop words as part of the chainValues
```go
func (c LLMChain) Call(values ChainValues) (ChainValues, error) {
	var stop []string
	promptValue, err := c.prompt.FormatPromptValue(values)
	if err != nil {
		return ChainValues{}, err
	}

	if stopVal, ok := values["stop"].([]string); ok {
		stop = stopVal
	}

	generations, err := c.llm.Generate([]string{promptValue.String()}, stop)
	if err != nil {
		return ChainValues{}, err
	}

	finalOutput, err := c.OutputParser.ParseWithPrompt(generations[0].Text, promptValue)
	if err != nil {
		return ChainValues{}, err
	}

	return ChainValues{c.OutputKey: finalOutput}, nil
}
```
# The secon one is as part of the Call()
```go
func (c LLMChain) Call(values ChainValues,stopWords []string) (ChainValues, error) {
	
	promptValue, err := c.prompt.FormatPromptValue(values)
	if err != nil {
		return ChainValues{}, err
	}


	generations, err := c.llm.Generate([]string{promptValue.String()}, stop)
	if err != nil {
		return ChainValues{}, err
	}

	finalOutput, err := c.OutputParser.ParseWithPrompt(generations[0].Text, promptValue)
	if err != nil {
		return ChainValues{}, err
	}

	return ChainValues{c.OutputKey: finalOutput}, nil
}
```

Abraxas-365:
Im using the Ts approach for now, let mw know wat are your thoughts

tmc:
Both the python and ts implementations prefer the former style so we should go with that.

Often in go we approach this with variadic option arguments so it'd look like .Call(values, chain.WithStopWords(words)) which I find more extensible, discoverable, and better-typed.

Let's match upstream for now until we decide we want to diverge.

Abraxas-365:
@tmc Okey allready pushed the changes with the first approach, pls review it when you have time and if you like it accept the PR pls

exp/chains/llmChain.go
```diff
@@ -32,14 +32,17 @@ func (c LLMChain) GetMemory() memory.Memory {
...
-	//TODO: stop
-
+	var stop []string
 	promptValue, err := c.prompt.FormatPromptValue(values)
 	if err != nil {
 		return ChainValues{}, err
 	}
 
-	generations, err := c.llm.Generate([]string{promptValue.String()})
+	if stopVal, ok := values["stop"].([]string); ok {
```
tmc, line 41: merits a comment

tmc (approved):
LGTM

PR #40, opened 2023-04-21
*: fix space in .github folder

Fraser-Isbester:
**summary:** there is a space in the .github root folder which can't be right.

closes https://github.com/tmc/langchaingo/issues/39


tmc:
Thanks for your contribution! this was causing some snags recently.

Issue #5, opened 2023-04-17
Add doc.go files to every package.


tmc:
Every package should have a well written package-level comment beginning with "// Package (packagename)".

teezzan:
@tmc, can I hop on this? Looks like something I can start with while getting to know the codebase for more contributions.

tmc:
> @tmc, can I hop on this? Looks like something I can start with while getting to know the codebase for more contributions.

@teezzan Yes please do!

PR #44, opened 2023-04-21
exp: change names of files and start using actual schema

FluffyKebab:
- Change file names in exp to follow go standard
 - Make exp dependent on the actual schema package
 - Add some comments in exp
 - Change from using reflect.DeepEqual to cmp.Equal in tests in exp
 - Change inputValues and chainValues to map[string]any in exp
 - Remove exp/schema
 - Modify memory interface in schema, changing SaveContext outputs from map[string]string to map[string]any
 - Add doc.go to packages in exp #5


tmc (approved):
LGTM

PR #56, opened 2023-04-24
chats: added openAi chat generation as LLM, fixes #43

### LINKED ISSUE #43: Introduce `chatModels`
I already started with openai but have the need to discuss how we want to map langchain to langchaingo in a few instances. I'll make a PR soon.

ChristianSch:
### PR Checklist

- [ ] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [ ] Describes source of new concepts.
- [ ] (remove and do not use this row unless you are a trusted contributor) I am an existing
  contributor.

### DIFF
--- a/chatmodels/openai/internal/openaiclient/chat.go
+++ b/chatmodels/openai/internal/openaiclient/chat.go
@@ -53,7 +53,7 @@ func (c *Client) createChat(ctx context.Context, payload *chatPayload) (*chatRes
 	}
 
 	body := bytes.NewReader(payloadBytes)
-	req, err := http.NewRequestWithContext(ctx, "POST", "https://api.openai.com/v1/chat/completions", body)
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, "https://api.openai.com/v1/chat/completions", body)
 	if err != nil {
 		return nil, err
 	}
--- a/chatmodels/openai/internal/openaiclient/chat.go
+++ b/chatmodels/openai/internal/openaiclient/chat.go
@@ -1,0 +1,79 @@
+package openaiclient
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"net/http"
+)
+
+// ChatPayloadMessage is a message in the format expected by the OpenAI API.
+type ChatPayloadMessage struct {
+	// Role is the role of the message, either "system", "user", or "assistant"
+	Role string `json:"role"`
+	// Content is the content of the message
+	Content string `json:"content"`
+}
+
+type chatPayload struct {
+	Model       string               `json:"model"`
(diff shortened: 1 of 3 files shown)


chatmodels/openai/internal/openaiclient/chat.go
```diff
@@ -0,0 +1,79 @@
...
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"net/http"
+)
+
+// ChatPayloadMessage is a message in the format expected by the OpenAI API.
+type ChatPayloadMessage struct {
```
ChristianSch, line 11: as already said in the comments in https://github.com/tmc/langchaingo/issues/43 this format is currently not compatible with the LLM interface but is required by the openAI API.

ChristianSch:
This, as is, not compatible with the internal API of this library. That's why this is a draft and doesn't include tests or anything. I would like to discuss how to proceed. 

To test the implementation, use the following main.go file:
https://gist.github.com/ChristianSch/20251608147883cd5c6752988f16eb8e

chatmodels/openai/openaichatllm.go
```diff
@@ -0,0 +1,68 @@
...
+
+	return []*llms.Generation{
+		{
+			Text:           result.Text,
+			GenerationInfo: map[string]any{},
+		},
+	}, nil
+}
+
+func New() (*LLM, error) {
```
ChristianSch, line 59: right now none of the LLMs are properly configurable by this contract. what I would like to do is something like this:

```go
type OpenAIChatConfig struct {
    Model       string
    // note the absence of "Messages" here, is provided in Call/Generate
    Temperature float64  `json:"temperature"`
    N           int      `json:"n"`
    MaxTokens   int      `json:"max_tokens"`
    Stop        []string `json:"stop,omitempty"` // could be moved to Call/Generate as well?
}

func New(config OpenAIChatConfig) (*LLM, error) {
    TODO()
}
```

ChristianSch: this would break with the rest of the LLMs but currently makes this lib not really usable imho

tmc: Let's do variadic/functional Options arguments to pass these (and have some reasonable defaults).

like

```go
func New(options ...Option) (*LLM, error) {
```

tmc (commented):
I don't think we should make a separate client here, instead we should move/refactor/extend the openai client used in the llm tree.

tmc:
I'm going to accept 56 and we can refactor this as needed.

tmc:
This is solved in other work/ merged PRs.

PR #60, opened 2023-04-25
memory: add version outside of exp

FluffyKebab:
- Add buffer memory
https://github.com/hwchase17/langchainjs/blob/main/langchain/src/memory/buffer_memory.ts
https://github.com/hwchase17/langchain/blob/master/langchain/memory/chat_memory.py
 - Add chat message history
https://github.com/hwchase17/langchainjs/blob/main/langchain/src/stores/message/in_memory.ts
https://github.com/hwchase17/langchain/blob/master/langchain/memory/chat_message_histories/in_memory.py

### PR Checklist


memory/buffer_memory.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// Clear sets the chat messages to a new and empty chat message history.
+func (m *Buffer) Clear() error {
+	m.ChatHistory.Clear()
+	return nil
+}
+
+func getInputValue(inputValues map[string]any, inputKey string) (string, error) {
+	if inputKey != "" {
```
tmc, line 101: this conditional seems unnecessary, could you explain in a comment?

memory/buffer_memory.go
```diff
@@ -0,0 +1,144 @@
...
+				ErrInvalidInputValues,
+				inputValues,
+				inputKey,
+			)
+		}
+
+		return getInputValueReturnToString(inputValue)
+	}
+
+	if len(inputValues) == 1 {
```
tmc, line 115: these branches should be commented

memory/buffer_memory.go
```diff
@@ -0,0 +1,144 @@
...
+
+// Buffer is a simple form of memory that remembers previous conversational back and forths directly.
+type Buffer struct {
+	ChatHistory *ChatMessageHistory
+
+	ReturnMessages bool
+	InputKey       string
+	OutputKey      string
+	HumanPrefix    string
+	AiPrefix       string
```
tmc, line 21: nit: Prefer AIPrefix

memory/empty_memory.go
```diff
@@ -0,0 +1,32 @@
+package memory
+
+import (
+	"github.com/tmc/langchaingo/schema"
+)
+
+// Empty memory is a class that implement the memory interface, but does nothing.
+// The class is used as default in multiple chains.
```
tmc, line 8: I see "SimpleMemory" in the python codebase but not an empty implementation -- should this be Simple vs Empty?

tmc (commented):
Looks pretty good, left some comments.

memory/empty_memory.go
FluffyKebab: Yes, empty was just a name I came up with. Didnt know somthing like it existed in the python or ts version. Empty output parser should also be renamed.

memory/buffer_memory.go
FluffyKebab: If the input key is set, the function will return the input key of the input values. Otherwise the function will return the only entry in input values and error if there are multiple or no keys. So the function has to check if the input key is set or not. But yeah, this should be better commented.

tmc (approved):
LGTM

PR #62, opened 2023-04-26
OpenAI model selection and Chat completions

rdbell:
### PR Checklist

- [ ] ~~Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.~~

This implements OpenAI model selection and chat completions.

It's implemented in a way that's functionally similar to the way we're already doing text completions. I understand there are some redesign discussions happening in https://github.com/tmc/langchaingo/pull/56 and https://github.com/tmc/langchaingo/issues/43. This has lots of overlap with https://github.com/tmc/langchaingo/pull/56. 

Needed this for some testing on another project but figured I'd share.

Feel free to cherry-pick or close this out entirely if there's nothing needed here.


llms/openai/openaillm.go
```diff
@@ -48,6 +52,16 @@ func (o *LLM) Generate(prompts []string, stopWords []string) ([]*llms.Generation
 	}, nil
 }
 
+// Chat requests a chat response for the given prompt.
+func (o *LLM) Chat(prompt string) (string, error) {
```
tmc, line 56: I'm fine with this but we should consider what interface to define that would satisfy this.

tmc (commented):
You're right that this overlaps a decent amount with https://github.com/tmc/langchaingo/pull/56 -- but I think this approach is what I prefer.

tmc (approved):
This looks good to me, we should define an interface that represents the Chat call (Chater?) but we can do that separately, I also want to make the refactor in #65 but that can also be separate.

PR #70, opened 2023-04-28
chain: add standard interface, run, call and llm 

FluffyKebab:
- Add interface for chains outside of exp. Maybe the call method in the interface should be renamed to avoid confiusion with the call function that should be used by users. 
 - Add run and call functions.
https://github.com/hwchase17/langchainjs/blob/main/langchain/src/chains/base.ts
https://github.com/hwchase17/langchain/blob/master/langchain/chains/base.py
 - Add llm chain.
 https://github.com/hwchase17/langchainjs/blob/main/langchain/src/chains/llm_chain.ts
https://github.com/hwchase17/langchain/blob/master/langchain/chains/llm.py

### PR Checklist


chain/chain.go
```diff
@@ -0,0 +1,122 @@
+package chain
```
tmc, line 1: this package should be "chains" to reflect upstream/python

chain/chain.go
```diff
@@ -0,0 +1,122 @@
...
+	ErrMultipleOutputsInRun = errors.New("run not supported in chain with more then one expected output")
+	// ErrMultipleOutputsInRun is returned in the run function if the chain returns
+	// a value that is not a string.
+	ErrWrongOutputTypeInRun = errors.New("run not supported in chain that returns value that is not string")
+)
+
+// Cain is the interface all chains must implement.
+type Chain interface {
+	// Call runs the logic of the chain and returns the output. This method should
+	// not be called directly. Use rather the Call function that handles the memory
```
tmc, line 32: hmm yeah this is a potentially confusing API, I'll put some thought into this.

chain/chain.go
```diff
@@ -0,0 +1,122 @@
...
+	ErrMultipleInputsInRun = errors.New("run not supported in chain with more then one expected input")
+	// ErrMultipleOutputsInRun is returned in the run function if the chain expects
+	// more then one output values.
+	ErrMultipleOutputsInRun = errors.New("run not supported in chain with more then one expected output")
+	// ErrMultipleOutputsInRun is returned in the run function if the chain returns
+	// a value that is not a string.
+	ErrWrongOutputTypeInRun = errors.New("run not supported in chain that returns value that is not string")
+)
+
+// Cain is the interface all chains must implement.
```
abhinavmaury, line 29: a typo here

tmc (approved):
I'll think about the documentation and how/if we can clear this up in terms of naming but I'd prefer this get in sooner than later. LGTM.

Issue #61, opened 2023-04-26
Feature Parity Matrix


tmc:
We should take inspiration (and content) from https://langchain.com/features.html and have a markdown file that indicates the current implementation state.

Once we have docs site (#21) this content could go there.

rdbell:
**LangChain Go Features**

**Feature** | **Status**
-- | --
Prompt Templates | ✅
Few Shot Prompt Template | ✅
Output Parsers | ✅
Example Selectors | ✅

**Text Splitters**  
**Feature** | **Status**
-- | --
Character Text Splitter | ✅
Recursive Character Text Splitter | ✅
Markdown Text Splitter | ✅

**Chains**  
**Feature** | **Status**
-- | --
LLM Chain | ✅
Stuff Combine Documents Chain | ✅
Map Reduce Combine Documents Chain | ✅
Refine Combine Documents Chain | ✅
Map Rerank Combine Documents Chain | ❌
Chat Vector DB Chain | ✅
Vector DB QA Chain | ✅
Analyze Document Chain | ✅
Question Answering Chains | ✅
Summarization Chains | ✅
Question Answering With Sources Chains | ❌
SQL Database Chain | ✅
API Chain | ❌
Transformation Chain | ❌
Constitutional Chain | ❌
Conversational Chain | ✅
Graph QA Chain | ❌
HyDE Chain | ❌
LLM Bash Chain | ❌
LLM Math Chain | ❌
PAL Chain | ❌
LLM Requests Chain | ❌
Moderation Chain | ❌
Sequential Chain | ❌
Simple Sequential Chain | ❌

**Agents**
**Feature** | **Status**
-- | --
zero-shot-react-description | ✅
chat-zero-shot-react-description | ✅
self-ask-with-search | ❌
react-docstore | ❌
conversational-react-description | ❌
chat-conversational-react-description | ✅

**Memory**
**Feature** | **Status**
-- | --
Buffer Memory | ✅
Buffer Window Meory | ✅
Summary Memory | ❌
Entity Memory | ❌
Summary Buffer Memory | ❌
Knowledge Graph Memory | ❌

tmc:
Awesome thank you @rdbell -- can you open a PR with this content in a markdown file? (you can just place it in ./docs in the root for now).

Also, please join the #langchain-go Discord channel in the langchain discord.

MarkEdmondson1234:
[...]

PR #73, opened 2023-04-28
docs: add docs folder and parity matrix md file

O-Roma:
Add a docs folder with the parity matrix md file from  #61

### DIFF
--- a/docs/parity_matrix.md
+++ b/docs/parity_matrix.md
@@ -7,7 +7,7 @@ Please note that this page lists the current state of the LangChain Go project, 
 ## Prompt Templates
 | Feature                            | Status |
 |------------------------------------|--------|
-| Prompt Templates                   | ❌     |
+| Prompt Templates                   | ✅     |
 | Few Shot Prompt Template           | ❌     |
 | Output Parsers                     | ❌     |
 | Example Selectors                  | ❌     |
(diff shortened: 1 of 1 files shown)


FluffyKebab:
The matrix is wrong. I think it is just copied from the typescript version.

In text splitters we don't have:
- Character Text Splitter
- Markdown Text Splitter

In chains we only have
- llm chain
- Stuff Combine Documents Chain
- And one question answering chain

In agents, I belive we have just the zero-shot-react-description one and in memory we only have the buffer memory currently. I also think the matrix should contain information about output parsers and document loaders.

O-Roma:
You are right @FluffyKebab . Sorry I didn't check. I got it from the issue and thought it was already checked. I updated the table with the info you provided. Now we should keep polishing it until it makes sense and it's not just a copy from the typescript version 😞

FluffyKebab:
@O-Roma No reason to be sorry, it was completely fair to assume that the matrix from the issue was correct.

tmc:
@O-Roma this looks ready to review, going to mark it as such.

tmc (approved):
Thanks! LGTM -- we'll be treating this as a living document so I'm going to go ahead an merge even if there's a thing or two to tweak.

PR #77, opened 2023-05-03
Logging concept

rdbell:
- Added `logger` package to give output on what the LLM is doing as an agent is running through its chain.
- Minor code cleanup/refactoring.

This replaces some basic `fmt.Println` statements and adds more verbose logging with better context and a visual system for viewing and debugging an LLM agent's process as it works through the chain of commands.

Here's what it looks like in action:

Screenshot 2023-05-03 at 1 21 43 AM

And in compact mode:

Screenshot 2023-05-03 at 1 23 32 AM

### PR Checklist

- [X] Name your Pull Request title clearly, concisely, and prefixed with the name of primarily affected package you changed according to [Good commit messages](https://go.dev/doc/contribute#commit_messages) (such as `memory: add interfaces for X, Y` or `util: add whizzbang helpers`).
- [X] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [X] Describes source of new concepts.
- [X] Contains test coverage for new functions.
- [X] Passes all [`golangci-lint`](https://golangci-lint.run/) checks.
- [X] I am an existing contributor.


tmc (commented):
I like this a lot but I think we should adjust the approach a bit: expose a Logger interface on this package and a default implementation like you have and then we can implement WithLogger() style options to enable passing around /setting loggers.

Thoughts?

bdqfork:
May be introduce callbacks instead? #108

navicstein:
I think this is super useful

tmc:
As we have callbacks landed (with a sample handler for logging) I thinik we can close this.

PR #93, opened 2023-05-18
Feat: Parse OpenAI status code, error message

obitech:
This MR does several things (see commit messages for further infos):
* Remove `go.mod` and `go.sum` from all examples (it's not needed and has the upside that it always uses the most up-to-date version of the code, without running `go get`)
* Return invalid OpenAI status code, parse error message
* Simplify model selection on completion call
* Fixes completion call by passing all options down to the client

### PR Checklist

- [ ] (remove and do not use this row unless you are a trusted contributor) I am an existing
  contributor.


tmc (commented):
Thanks, can you undo the changes to the examples? I want them individually compilable and we have some make targets to handle updating them. Open to changes here but let's discuss that separately.

tmc (approved):
There are some other aspects of the errors: param, and code, but this is a good improvement.

tmc (approved):
LGTM

PR #95, opened 2023-05-19
agents: refactor to better match upstream and increase consistency

FluffyKebab:
This refactors the agents and tool packages to better match the typescript and python version and increase consistency with other packages in this version. 

In the current agent package all of the logic for executing the mrkl agent is a part of the mrkl agent. This is not how it is done upstream. In both the typescript and python version there is a executor that inherits from the base chain that does the agent steps and runs the tools. The agent classes are also stateless and take previous steps and the inputs and returns the action of the agent. This therefore moves the execution logic of the mrkl agent to an executor chain in the executor package.

https://github.com/hwchase17/langchainjs/blob/main/langchain/src/agents/executor.ts
https://github.com/hwchase17/langchain/blob/master/langchain/agents/agent.py line 620

Instead of implementing the agent executor interface the mrkl agent then implements the agent interface. 

```go
// Agent is the interface all agents must implement.
type Agent interface {
	// Given an input and previous steps decide what to do next. Returns either actions or a finish.
	Plan(ctx context.Context, intermediateSteps []schema.AgentStep, inputs map[string]string) ([]schema.AgentAction, *schema.AgentFinish, error) //nolint:lll
	GetInputKeys() []string
	GetOutputKeys() []string
}
```

In both the typescript and python version there is different classes for agents doing multiple steps and single steps and a class called agent that inherits from the agent class doing a single step, but I dont think having multiple interfaces for the agents will work well in a Go version. 

The function creating an executor is moved to the executor package and is renamed to "Initialize" to match upstream.

Tool is refactored to be a interface instead of a struct to better match other packages and a calculator tool is also added.

A test running the mrkl agent executor and a test in the mrkl package is added.

### PR Checklist


exp/tools/calculator.go
```diff
@@ -0,0 +1,33 @@
...
+func (c Calculator) Name() string {
+	return "calculator"
+}
+
+// Call evaluates the input using a starlak evaluator and returns the
+// result as a string.
+func (c Calculator) Call(input string) (string, error) {
+	v, err := starlark.Eval(&starlark.Thread{Name: "main"}, "input", input, math.Module.Members)
+	if err != nil {
+		return "I don't know how to do that.", nil
```
tmc, line 29: this kind of hides the error state

tmc (approved):
Thanks, this is a nice set of improvements. -> Let's get this aimed to land outside of the exp dir?

PR #107, opened 2023-05-28
chat message history: introduce interface type, add new history package

mvrilo:
- Introducing a `ChatMessageHistory` interface:
```go
type ChatMessageHistory interface {
    Messages() ([]ChatMessage, error)
    AddMessage(ChatMessage) error
    Clear() error
}
```

- Renamed `ChatMessageHistory` to `SimpleChatMessageHistory` as in-memory implementation
- Added one more example llm chain example using buffer and history

References:
https://github.com/hwchase17/langchainjs/tree/main/langchain/src/stores/message
https://github.com/hwchase17/langchain/tree/master/langchain/memory/chat_message_histories

### PR Checklist


memory/history/simple.go
```diff
@@ -0,0 +1,55 @@
+package history
+
+import "github.com/tmc/langchaingo/schema"
+
+// Statically assert that the types implement the interface.
+var _ schema.ChatMessageHistory = (*SimpleChatMessageHistory)(nil)
+
+// SimpleChatMessageHistory is a struct that stores chat messages.
+type SimpleChatMessageHistory struct {
```
alvarowolfx, line 9: the `SimpleChatMessageHistory` is the equivalent to the original In-memory implementation. Should we just call it InMemoryChatMessageHistory instead ?

alvarowolfx (commented):
Added a comment about naming the In Memory implementation. Can we add this history memory implementation to one of the chat samples/examples ?

memory/history/simple.go
mvrilo: I kept the "simple" prefix because the in-memory `Memory` implementation is also referenced as "simple" here:
https://github.com/tmc/langchaingo/blob/main/memory/simple.go#L9

mvrilo:
By default the [`Buffer` memory impl](https://github.com/mvrilo/langchaingo/blob/history-interface/memory/buffer.go#L32) already uses the `SimpleChatMessageHistory`.

Here's a quick usage, setting it manually:

```
mem := memory.NewBuffer()
mem.ChatHistory = history.NewSimpleChatMessageHistory()

llmChain := chains.NewLLMChain(llm, prompts.NewPromptTemplate("How old is the planet {{.planet}}?", []string{"planet"}))
llmChain.Memory = mem

ctx := context.Background()
out, err := chains.Call(ctx, llmChain, map[string]any{"planet": "Earth"})
if err != nil {
    return err
}

fmt.Println(out["text"])
```

I wanted to keep the PR small but I can add this in the examples, no problem.

Another relevant change that I could add to this PR, before adding the example, is changing the `buffer.New()` to `buffer.New(history schema.ChatMessageHistory)` in the `Buffer` impl.

mvrilo:
Updated my PR changing interface methods to return errors, added one more example and updated the `Buffer` initializer to `buffer.New(schema.ChatMessageHistory)`.

tmc:
Why not just accept the general ChatMessage interface vs the more specific methods here?

mvrilo:
Where exactly @tmc? You mean for the `buffer.NewBuffer` to accept a `[]schema.ChatMessage` instead of the new type?

tmc:
> Where exactly @tmc? You mean for the `buffer.NewBuffer` to accept a `[]schema.ChatMessage` instead of the new type?

Sorry I missed this message. 

I mean why not something like:

```go
type ChatMessageHistory interface {
    Messages() ([]ChatMessage, error)
    AddMessage(ChatMessage) error
    Clear() error // note: I'm not sure clear is necessary here.
}
```

mvrilo:
Thanks! The other implementations had the AddMessage for each role, that's why I kept that way. I'll update the PR.

tobiade:
Actually, I just realised I had a brain-fart moment and could have just done:
```go
rc := textsplitter.NewRecursiveCharacter()
rc.ChunkSize = 1024
rc.ChunkOverlap = 0
```

I'm happy to close this if you think functional options isn't that important here.

PR #109, opened 2023-05-28
Improving HuggingfaceLLM: Configurable RepoID option

EvilFreelancer:
### PR Checklist

---

Phew, your questionnaire is quite complex, feels like I'm taking exams :)

Here's a brief description of this PR. It allows the use of HuggingFace pipelines by specifying a custom repository instead of hardcoded "google/flan-t5-xl" (which not work btw). For my tests, I used the GPT-2 model, but any other model can be used if desired.

I implemented the mentioned setting through `llms.WithModel` and demonstrated its usage in the corresponding example. I'm sure this will help newcomers like me to understand how to use this library more quickly.

Thank you guys for this amazing extension! I've trying to integrate executing of Python scripts to LocalAI and you saved a lot of my time.

PS. Idk how to test it correctly, al lot of existing tests don't check similar thing, so maybe you may help me with it.

### DIFF
--- a/examples/huggingface-llm-example/huggingface_example.go
+++ b/examples/huggingface-llm-example/huggingface_example.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 
+	"github.com/tmc/langchaingo/llms"
 	"github.com/tmc/langchaingo/llms/huggingface"
 )
 
@@ -14,7 +15,11 @@ func main() {
 		log.Fatal(err)
 	}
 	ctx := context.Background()
-	completion, err := llm.Call(ctx, "What would be a good company name be for name a company that makes colorful socks?")
+
+	var generateOptions []llms.CallOption
+	generateOptions = append(generateOptions, llms.WithRepoId("gpt2"))
+
+	completion, err := llm.Call(ctx, "What would be a good company name be for name a company that makes colorful socks?", generateOptions...)
 	if err != nil {
 		log.Fatal(err)
 	}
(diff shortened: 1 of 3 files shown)


tmc (approved):
Thanks, that was nagging me, LGTM.

tmc:
"Phew, your questionnaire is quite complex, feels like I'm taking exams :)" -- you're right that this is probably too onerous -- can you open a PR to shorten this checklist to the essentials?

tmc:
@EvilFreelancer After thinking a bit more -> why can't we use/re-use the WIthModel option for this?

EvilFreelancer:
> "Phew, your questionnaire is quite complex, feels like I'm taking exams :)" -- you're right that this is probably too onerous -- can you open a PR to shorten this checklist to the essentials?

Hi! You mean just update the description or recreate PR from scratch? (close this and create another one)

EvilFreelancer:
> @EvilFreelancer After thinking a bit more -> why can't we use/re-use the WIthModel option for this?

I've just noticed this function, didnt saw it earlier, but yeah, why not :) ill change to it then

EvilFreelancer:
Now it looks better, thanks for the tip!

tmc (approved):
this will probably blow up if opts.model is empty -- I can accept this but do you think we should have a default?

EvilFreelancer:
> this will probably blow up if opts.model is empty -- I can accept this but do you think we should have a default?

It's a good idea, but id model will be an empty user will see an error about this,

PR #110, opened 2023-05-29
pinecone similaritySearch: Include score threshold based filtering

cduggn:
### PR Checklist

## Description
This PR will allow similarity searches to optionally include threshold-based filtering. A minimum threshold between 0 and 1 can be set using a newly added `vectorstores.WithScoreThreshold` function. This function sets the `ScoreThreshold` option. This defines a minimum score threshold indicating to the SimilaritySearch function to only return documents with a score that meets or exceeds that threshold. 
- This will allow filtering of less relevant or lower quality results. 
- Similar functionality exists in Python langchain in the form of the [similarity_search_with_relevance_score](https://github.com/hwchase17/langchain/blob/master/langchain/vectorstores/base.py). In that version `similarity_search_with_relevance_score` is a separate function which is called by `similarity_search`, filtering only happens if the similarity argument is used.

I implemented the changes in `pinecone.SimilaritySearch` , `pinecone.restQuery` . 
- If the `ScoreThreshold` property is not explicitly set it will assume the default value for a float64 (0) which will match all returned documents.  
- When `restQuery` executes it checks if the score threshold is a value that is not 0 (the default), this implies the threshold was set. Logic then filters based on that threshold. 
- The threshold can only be a value between 0 and 1. This range is handled in "pinecone.getScoreThreshold" and throws an error if the value falls outside the range.  
- I've added the change only for the `restQuery` function (and not `grpcQuery` although if the change is accepted it could be applied with minimal changes). The restQuery


tmc (approved):
LGTM thanks

tmc (approved):
LGTM thanks

tmc:
hm, lint failure. I'm a bit on the fence wrt gofumpt.. will fix.

note: I've updated the repo settings to allow ci to run on initial PRs.

cduggn:
Thanks @tmc. I can take a look at applying same change to `grpcQuery`.

PR #128, opened 2023-06-14
llms: split LLMChat and LLM into different structs

FluffyKebab:
I saw in the upstream versions that the there are different classes for the chat models and standard llms. Doing something similar in the go version would bring multiple benefits. Better prompt selection being one example. This therefore changes the chat interface to have a  call and a generate function, like the llm interface, and splits the llm structs into a chat and llm struct. Both the openai llm and chat struct implement the LanguageModel interface. To do this the chat generation and generation types is united into one type, and functionality for token counting is added. 

I also saw that some types are defined in both the schema package and the llm package. For example the generation type. What versions should be removed? I think that it would be best to have the types used for llms in the llms package.  

### References
https://github.com/hwchase17/langchain/blob/master/langchain/base_language.py
https://github.com/hwchase17/langchainjs/blob/main/langchain/src/base_language/index.ts 

### PR Checklist


tmc (approved):
I like this largely and am open to thinking shout the right home for those types. Let’s get this in and consider iteration.

PR #129, opened 2023-06-16
pinecone restquery: Add filter logic to limit searches based on metadata

cduggn:
### PR Checklist

- [x]  Read the [Contributing documentation](https://github.com/tmc/langchaingo/blob/main/CONTRIBUTING.md).
### Description
This PR will allow the Pinecone similarity search to include metadata filtering to limit vector searches based on some specified metadata key-value pairs. This is functionality offered by the Pinecone query API. Using metadata filters the search queries will retrieve exactly the number of nearest-neighbor results that match the filters. Metadata filtering can lower the search latency when compared with unfiltered searches. The built in Pinecone [metadata filtering language](https://docs.pinecone.io/docs/metadata-filtering) provides the tools to build more complex search queries.  

#### Sample usage 
Metadata filtering allows for more complex search queries. This query reduces the vector search to only include matches located in the kitchen or sitting room, and with a square footage of greater than 300. 
```golang
	filter := map[string]interface{}{
		"$and": []map[string]interface{}{
			{
				"location": map[string]interface{}{
					"$in": []string{"office", "kitchen", "sitting room"},
				},
			},
			{
				"square_feet": map[string]interface{}{
					"$gte": 300,
				},
			},
		},
	}

	result, err := chains.Run(
		context.TODO(),
		chains.NewRetrievalQAFromLLM(
			llm,
			vectorstores.ToRetriever(store, 5, vectorstores.WithNameSpace(
				id), vectorstores.WithFilters(filter)),
		),
		"What color is the lamp in each room?",
	)
```

### Existing implementation
Metadata filtering is implemented for Pinecone similarity searches in upstream versions of the library including [Python Langchain](https://github.com/hwchase17/langchain/blob/master/langchain/vectorstores/pinecone.py)

### Changes
Changes for this feature were implemented in `options.go`, `restapi.go` and `pinecone.go`

-  A new `WithFilters` option in `options.go` which take a parameter of type `map[string]any`. The `any` type is used to support each of the [metadata types](https://docs.pinecone.io/docs/metadata-filtering) allowed by the pinecone API.
- `getFilters` function was added to `pinecone` package. This package also includes changes to `AddDocuments` and `SimilaritySearch` functions. The `getOptions` function is moved into `SimilaritySearch` to reduce the frequency with which it will be called as new options are added.  The `restQuery` function call is updated to include the filters parameter. 
- `restQuery` payload updated to include a new filter type. If the filter type is not set the value will default to nil. Pinecone will ignore this option.  

** Note: 
- Invalid filter requests will be handled by the Pinecone API and will result in a compilation error. There is not additional error handling or validation carried out on the Filter type before being sent to Pinecone. The filter option in Pinecone can be constructed in a number of ways using strings, lists, booleans and numbers making it complex to accurately validate. 
- Pinecone recommends queries with top_k over 1000 should not contain `include_metadata=True` or `include_data=True`


tmc (approved):
LGTM

PR #136, opened 2023-06-19
prompts: add chat prompt template

### LINKED ISSUE #134: [Feature Request] Introducing MessagePromptTemplate
First of all, please correct me if I am mistaken about your roadmap plan.

### Your proposal is related to a problem

Currently, I couldn't find a proper way to invoke `schema.ChatMessage` with the package `prompts`. The only related one I found is `prompts.StringPromptValue` which Format the message to `schema.HumanChatMessage`.

The original Langchain provides different types of MessagePromptTemplate such as `AIMessagePromptTemplate`, `HumanMessagePromptTemplate`, and `SystemMessagePromptTemplate`.

FYI: https://langchain-langchain.vercel.app/docs/modules/model_io/prompts/prompt_templates/msg_prompt_templates

This is more convenient instead of using schema.ChatMessage directly.

### Describe the solution you'd like

So let's introduce the same thing into this lib.

To declare an interface BaseMessagePromptTemplate inside `prompts`:
```go
type BaseMessagePromptTemplate interface {
	IntputVariables() []string
	FormatPrompt(values map[string]interface{}) (schema.PromptValue, error)
	FormatMessages(values map[string]interface{}) ([]schema.ChatMessage, error)
}
```
Then we implement `AIMessagePromptTemplate`, `HumanMessagePromptTemplate`, and `SystemMessagePromptTemplate` based on `prompts.PromptTemplate` for the interface. That tree prompt templates should produce their respective `schema.PromptValue` (`AIPromptValue`, `HumanPromptValue`, and `SystemPromptValue`) so that can be easily used by Chat Models in such way:
```go
func TestChatWithTemplate(t *testing.T) {
	template := prompts.NewChatPromptTemplate([]prompts.BaseMessagePromptTemplate{
		chat.NewSystemMessagePromptTemplate(
			`You are a translation engine that can only translate text and cannot interpret it.`,
			nil,
		),
		chat.NewHumanMessagePromptTemplate(
			`translate this text from {{.input_lang}} to {{.output_lang}}:\n{{.input}}`,
			[]string{"input_lang", "output_lang", "input"},
		),
	})
	value, err := template.FormatPrompt(map[string]interface{}{
		"input_lang":  "English",
		"output_lang": "Chinese",
		"input":       INPUT,
	})
	if err != nil {
		t.Fatalf("%s", err)
	}
	completion, err := llm.Chat(context.Background(), value.Messages())
	if err != nil {
		t.Fatalf("%s", err)
	}
	fmt.Println(completion.Message.GetText())
}
```

### Additional context

I would like to create a **PR** and make a contribution to this feature request if you let me know.:)

FluffyKebab:
This adds chat prompt templates and prompt templates for each of the message types. All message type prompt templates implement the new message formatter interface. GetInputVariables is added to the format prompter because the llm chain needs to know what input values the prompt needs. The checks for "stop" is removed because stop values are now given in the call options. A prompt value for chat messages is added.

Fixes #134. 

### PR Checklist

### DIFF
--- a/prompts/chat_prompt.go
+++ b/prompts/chat_prompt.go
@@ -1,0 +1,26 @@
+package prompts
+
+import (
+	"fmt"
+
+	"github.com/tmc/langchaingo/schema"
+)
+
+var _ schema.PromptValue = ChatPromptValue{}
+
+// ChatPromptValue is a prompt value that is a list of chat messages.
+type ChatPromptValue []schema.ChatMessage
+
+// String returns the chat message slice as a buffer string.
+func (v ChatPromptValue) String() string {
+	s, err := schema.GetBufferString(v, "Human", "AI")
+	if err == nil {
+		return s
+	}
+	return fmt.Sprintf("%v", []schema.ChatMessage(v))
+}
+
+// Messages returns the ChatMessage slice.
+func (v ChatPromptValue) Messages() []schema.ChatMessage {
+	return []schema.ChatMessage(v)
+}
--- a/prompts/chat_prompt_template.go
+++ b/prompts/chat_prompt_template.go
@@ -19,7 +19,7 @@ var (
 )
 
 // FormatPrompt formats the messages into a chat prompt value.
-func (p ChatPromptTemplate) FormatPrompt(values map[string]any) (schema.PromptValue, error) {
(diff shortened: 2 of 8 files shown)


tmc (approved):
LGTM

PR #148, opened 2023-06-25
chains: use langauge model instead of LLM and add predict 

FluffyKebab:
This changes the chains and agents packages to use the language model interface instead of the LLM interface so that chat models can be used. The predict function is also added to execute chains that returns a single string. The function exists in the python and typescript version.
https://github.com/hwchase17/langchain/blob/afc292e58d768d96245272ebe7d985f72b03da73/langchain/chains/llm.py

### PR Checklist


tmc (approved):
LGTM

PR #164, opened 2023-07-02
textsplitter: allow configuration of RecursiveCharacter

tobiade:
### Description
Think it'll be helpful to allow clients to configure text splitters while keeping some of the sane defaults in place. I recently had to configure `ChunkSize` and `ChunkOverlap` but couldn't do so without manually creating a `RecursiveCharacter`:
```go
rc := textsplitter.RecursiveCharacter{
		Separators:   []string{"\n\n", "\n", " ", ""},
		ChunkSize:    1024,
		ChunkOverlap: 0,
}
```

I've tried to follow the patterns we have for functional options in other packages.

### PR Checklist

- [x ] Read the [Code of conduct documentation](https://github.com/tmc/langchaingo/blob/main/CODE_OF_CONDUCT.md).

### DIFF
--- a/textsplitter/options.go
+++ b/textsplitter/options.go
@@ -1,0 +1,32 @@
+package textsplitter
+
+// Option is a function that configures Options.
+type Option func(*Options)
+
+// Options is a set of options for configuring a text splitter.
+type Options struct {
+	Separators   []string
+	ChunkSize    int
+	ChunkOverlap int
+}
+
+// WithSeparators is an option for setting the separators.
+func WithSeparators(separators []string) Option {
+	return func(o *Options) {
+		o.Separators = separators
+	}
+}
+
+// WithChunkSize is an option for setting the chunk size.
+func WithChunkSize(size int) Option {
+	return func(o *Options) {
+		o.ChunkSize = size
+	}
+}
+
+// WithChunkSize is an option for setting the chunk overlap.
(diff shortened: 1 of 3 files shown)


textsplitter/options.go
```diff
@@ -0,0 +1,32 @@
+package textsplitter
+
+// Option is a function that configures Options.
+type Option func(*Options)
+
+// Options is a set of options for configuring a text splitter.
+type Options struct {
```
tobiade, line 7: Please let me know if this needs to be an option that's specific to `RecursiveCharacter` and not a package-wide option.

tobiade: I'm also happy to un-export the fields in `RecursiveCharacter` if needed, given they'll be configured by the options.

tobiade:
Actually, I just realised I had a brain-fart moment and could have just done:
```go
rc := textsplitter.NewRecursiveCharacter()
rc.ChunkSize = 1024
rc.ChunkOverlap = 0
```

I'm happy to close this if you think functional options isn't that important here.

tmc:
I'm not sure this circumstance warrants functional arguments -> I generally want to reserve those for the configuration of key clients and methods that are going to be used in many cases and should have a dead simple set of defaults. I think specifying a/the RecursiveCharacter struct literal is fine here

tobiade:
Sure, makes sense.

PR #169, opened 2023-07-04
tools/duckduckgo: fix for response status 403 and 302

mvrilo:
- Adding a `user-agent` fixes `403` response status which seems to happen randomly.
- Adding a subdomain `html` to the search url to avoid a redirect (`302` response status).
- Also: accepting the `maxResults` and `userAgent` via the `New` func.

You can test the redirects with curl:

```
curl -sf -o /dev/null -D - 'https://duckduckgo.com/html/?q=olivia%20wilde' | grep 'HTTP\|ocation'
HTTP/2 302
location: https://html.duckduckgo.com/html/?q=olivia%20wilde
```
vs
```
curl -sf -o /dev/null -D - 'https://html.duckduckgo.com/html/?q=olivia%20wilde' | grep 'HTTP\|ocation'
HTTP/2 200
```

### PR Checklist

### DIFF
--- a/tools/duckduckgo/ddg.go
+++ b/tools/duckduckgo/ddg.go
@@ -14,10 +14,9 @@ type Tool struct {
 
 var _ tools.Tool = Tool{}
 
-func New() (*Tool, error) {
-	maxResults := 5
+func New(maxResults int, userAgent string) (*Tool, error) {
 	return &Tool{
-		client: internal.New(maxResults),
+		client: internal.New(maxResults, userAgent),
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


tools/duckduckgo/ddg.go
```diff
@@ -14,10 +14,9 @@ type Tool struct {
 
 var _ tools.Tool = Tool{}
 
-func New() (*Tool, error) {
-	maxResults := 5
+func New(maxResults int, userAgent string) (*Tool, error) {
```
tmc, line 17: should be documented.

it doesn't seem all that important to supply a user agent - can you export a default user agent so that folks can reach for it?

tmc (approved):
please follow up with suggestions

PR #170, opened 2023-07-04
huggingface: build inferencePayload appropriately

tobiade:
### Description
When using the Hugging Face client, parameters are not being correctly included in the `inferencePayload`, causing the Inference API to ignore them. The [docs](https://huggingface.co/docs/api-inference/detailed_parameters#text-generation-task) for performing a text generation task specify that the parameters should be specified in the body of the request like so:
```json
{
    "inputs": "Can you please let us know more details about your ",
    "parameters": {
        "top_k": 10
    }
}
```

but we're currently doing this:
```json
{
    "inputs": "Can you please let us know more details about your ",
    "top_k": 10
}
```

I've been able to test locally that this proposed change works.

### PR Checklist

### DIFF
--- a/llms/huggingface/internal/huggingfaceclient/huggingfaceclient.go
+++ b/llms/huggingface/internal/huggingfaceclient/huggingfaceclient.go
@@ -14,6 +14,7 @@ var (
 type Client struct {
 	Token string
 	Model string
+	url   string
 }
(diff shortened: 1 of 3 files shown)


tmc (approved):
Thank you.

PR #175, opened 2023-07-07
chains: add map reduce documents and apply

FluffyKebab:
This adds the map reduce document combing chain and the apply function to run a chain with multiple inputs at the same time. 
https://github.com/hwchase17/langchainjs/blob/main/langchain/src/chains/combine_docs_chain.ts
https://github.com/hwchase17/langchainjs/blob/main/langchain/src/chains/base.ts

### PR Checklist


chains/chains.go
```diff
@@ -121,6 +121,53 @@ func Predict(ctx context.Context, c Chain, inputValues map[string]any, options .
...
+
+	inputJobs := make(chan map[string]any, len(inputValues))
+	resultsChan := make(chan struct {
+		result map[string]any
+		err    error
+	}, len(inputValues))
+	defer close(inputJobs)
+	defer close(resultsChan)
+
+	for w := 0; w < maxWorkers; w++ {
```
tmc, line 140: can you double check this doesn't suffer from the go gotcha https://github.com/golang/go/issues/60078

tmc (approved):
great work!

chains/chains.go
FluffyKebab: I dont think it does. The variable w is not used in the anonymous function.

chains/chains.go
```diff
@@ -121,6 +121,53 @@ func Predict(ctx context.Context, c Chain, inputValues map[string]any, options .
...
+	}
+
+	results := make([]map[string]any, len(inputValues))
+	for i := range inputValues {
+		r := <-resultsChan
+		if r.err != nil {
+			return nil, r.err
+		}
+		results[i] = r.result
+	}
```
tobiade, line 166: How do we know what output corresponds to a particular input? This seems to be storing results in random order. I think we should be passing the index of the input when we write to `inputJobs`.

In `mapResultsToLLMChainInputs` I think you could possibly be mapping wrong results to inputs.

I'm also assuming this `Apply` function can be used for things other than MapReduce, so we'd need to know the output for a particular input.

I could be missing something though - thoughts?

FluffyKebab: Yes you are correct. I will make an update for this.

chains/chains.go
```diff
@@ -121,6 +121,53 @@ func Predict(ctx context.Context, c Chain, inputValues map[string]any, options .
...
+	if maxWorkers <= 0 {
+		maxWorkers = _defaultApplyMaxNumberWorkers
+	}
+
+	inputJobs := make(chan map[string]any, len(inputValues))
+	resultsChan := make(chan struct {
+		result map[string]any
+		err    error
+	}, len(inputValues))
+	defer close(inputJobs)
```
tobiade, line 137: Nitpick: we can close the `inputJobs` channel immediately after writing inputs to it so we can release the consuming goroutines before the `Apply` method finishes.

chains/chains.go
```diff
@@ -121,6 +121,53 @@ func Predict(ctx context.Context, c Chain, inputValues map[string]any, options .
 	return outputValue, nil
 }
 
+const _defaultApplyMaxNumberWorkers = 5
+
+// Apply executes the chain for each of the inputs asynchronously.
+func Apply(ctx context.Context, c Chain, inputValues []map[string]any, maxWorkers int, options ...ChainCallOption) ([]map[string]any, error) { // nolint:lll
```
tobiade, line 127: One for a different PR, but it'd be good to enable timeouts/cancellations using the `ctx` so chains don't run indefinitely (if that ever happens).

FluffyKebab: @tobiade can you check out #178?

tobiade: Thanks @FluffyKebab - left some comments.

PR #178, opened 2023-07-07
chains: add QA and summarization using map reduce

FluffyKebab:
This fixes some problems with the input keys in the map reduce chain and adds QA and summarization using the map reduce document chain. The apply function is also updated such that outputs and inputs are at the same index. 

### PR Checklist


chains/chains.go
```diff
@@ -123,46 +124,82 @@ func Predict(ctx context.Context, c Chain, inputValues map[string]any, options .
...
-					result map[string]any
-					err    error
-				}{
-					result: res,
-					err:    err,
+			defer wg.Done()
+			for {
+				select {
+				case <-ctx.Done():
+					return
```
tobiade, line 156: We can still write to `resultsChan` here with the error returned by `ctx.Err` i.e.
```go
resultsChan <- applyResult{
    err: ctx.Err()
    // can add a sentinel value for `i` e.g. -1 if needed, but don't think it's that important atm
}
```
That enables the client to know if the `Apply` function failed due to a timeout/cancellation.

tobiade: Ignore - I've just seen `getApplyResults` 😄

chains/chains_test.go
```diff
@@ -38,3 +41,24 @@ func TestApply(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, numInputs, len(results), "number of inputs and results not equal")
 }
+
+func TestApplyWithCanceledContext(t *testing.T) {
```
tobiade, line 45: We should uplift the `TestApply` function to be more comprehensive - it should test that an expected output exists for a given input.

chains/chains.go
tmc: There’s also cancel reason in newer go releases.

chains/chains_test.go
```diff
@@ -13,6 +15,7 @@ import (
 type testLanguageModel struct{}
 
 func (l testLanguageModel) GeneratePrompt(_ context.Context, _ []schema.PromptValue, _ ...llms.CallOption) (llms.LLMResult, error) { //nolint:lll
+	time.Sleep(time.Second)
```
tmc, line 18: intentional?

tmc (approved):
LGTM

tmc (approved):
LGTM

chains/chains_test.go
FluffyKebab: It is not necessary. I can remove it.

Issue #68, opened 2023-04-27
feature: add Map-Rerank Combine Documents Chain


jodosha:
## Description

> This method involves running an initial prompt on each chunk of data, that not only tries to complete a task but also gives a score for how certain it is in its answer. The responses are then ranked according to this score, and the highest score is returned.

[[source](https://docs.langchain.com/docs/components/chains/index_related_chains#map-rerank)]

## Acceptance Criteria

* It must rank in parallel with the ranking for each chunk of data and return the result with the highest score.
* It must work similarly to "Map Reduce Combine Documents Chain" but perform fewer LLM calls.
* It shouldn't combine results from one document but return a response for each chunk of data.

## References

* Original Python implementation: https://github.com/hwchase17/langchain/pull/516
* Golang features: https://github.com/tmc/langchaingo/issues/61

PR #182, opened 2023-07-10
feature: map rerank chain

baoist:
#68.

Relevant Documents:
- [LangChain Map Rerank](https://docs.langchain.com/docs/components/chains/index_related_chains#map-rerank).
- [Python Implementation](https://github.com/hwchase17/langchain/pull/516).

Would appreciate sanity check here, not sure I fully grok this chain. I believe this behaves similarly to its python counterpart, with some functionality provided by @FluffyKebab's `MapReduce` addition.

Changes:
- Moves `_intermediateStepsOutputKey` var to central file (out of `chains/map_reduce.go`).
- Fixes the output of the `RegexParser` so it can be used by LLM chains, updates/adds some documentation.
- Adds MapRerankDocuments functionality.

### PR Checklist

### DIFF
--- a/chains/map_rerank_documents.go
+++ b/chains/map_rerank_documents.go
@@ -1,0 +1,190 @@
+package chains
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"strconv"
+
+	"github.com/tmc/langchaingo/memory"
+	"github.com/tmc/langchaingo/outputparser"
+	"github.com/tmc/langchaingo/schema"
+	"golang.org/x/exp/maps"
+)
+
+const (
+	_mapRerankDocumentsDefaultDocumentTemplate = "{{.page_content}}"
+	_mapRerankDocumentsDefaultRankKey          = "score"
+	_mapRerankDocumentsDefaultAnswerKey        = "answer"
+)
+
+type MapRerankDocuments struct {
+	// Chain used to rerank the documents.
+	LLMChain *LLMChain
+
+	// Number of workers to concurrently run apply on documents.
+	MaxConcurrentWorkers int
+
+	// The input variable where the documents should be placed int the LLMChain.
+	LLMChainInputVariableName string
+
+	// The name of the document variable in the LLMChain.
+	DocumentVariableName string
+	// Key used to access document inputs.
+	InputKey string
+	// Key used to access map results.
+	OutputKey string
+
+	// Key used for comparison to sort documents.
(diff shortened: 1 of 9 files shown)


chains/map_rerank_documents.go
```diff
@@ -0,0 +1,196 @@
...
+		}
+
+		outputs[i] = c.parseMapResults(rankedAnswer)
+	}
+
+	sort.Slice(outputs, func(i, j int) bool {
+		curr, err := strconv.Atoi(outputs[i][c.RankKey].(string))
+		if err != nil {
+			return false
+		}
```
tobiade, line 104: Is there a reasoning for ignoring the error here? Why not do the integer conversion and sorting separately so we can handle conversion errors appropriately?

baoist: > Is there a reason for ignoring the error here?

Thought is that if it cannot be converted that we can assume the "compare" map should take precedence over this in sorting without needing to continue. With the inverse being true below.

> Why not do the integer conversion and sorting separately so we can handle conversion errors appropriately?

I had considered this, but I've been trying for parity with the python counterpart, which doesn't convert the type in the output:
```
[{'answer': ' Leo', 'score': '100'}]
```

That being said, there's still some cleanup and functionality to improve/fix, and I don't have a strong preference.

tobiade: Can you point me to the piece of code you're talking about? 

Looks like the Python [source](https://github.com/hwchase17/langchain/blob/b96ac13f3d17b7b2c9295fe369ab81439c105aca/langchain/chains/combine_documents/map_rerank.py#L192) does attempt integer conversion and will throw an error if it can't perform the conversion.

baoist: Same source.

I'm fine on raising/returning on failed conversion, but noting the output isn't typed as an int:
```
(Pdb) sorted_res[0][0]['score']
'100'
(Pdb) type(sorted_res[0][0]['score'])
<class 'str'>
(Pdb) sorted_res[0][0]
{'answer': ' Leo', 'score': '100'}
```

tobiade: Ok I think I see where the disconnect is now. I meant we should do the integer conversion of `score` for the sake of sorting, but the output can still have `score` as a string (which is the Python counterpart's behaviour)

tmc (approved):
Awesome! Thanks!

Issue #114, opened 2023-06-07
Feature request: SimpleSequentialChain


Tochemey:
@tmc is there any plan to add SimpleSequentialChain?

zesagata:
@Tochemey  Do you already work on this ? i can help

Tochemey:
@zesagata Not yet

heiningair:
+1

tobiade:
I'll pick this up.

PR #183, opened 2023-07-10
chains: simplesequentialchain and sequentialchain implementation

tobiade:
### Description
* Implementation of `SimpleSequentialChain` and `SequentialChain`. Addresses #114.
* Will add examples once this is merged - looks like I can't add an example with this PR.
* [Python source](https://github.com/hwchase17/langchain/blob/master/langchain/chains/sequential.py)

### PR Checklist

### DIFF
--- a/chains/sequential.go
+++ b/chains/sequential.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/tmc/langchaingo/internal/util"
 	"github.com/tmc/langchaingo/memory"
 	"github.com/tmc/langchaingo/schema"
 )
@@ -32,11 +33,11 @@ func NewSequentialChain(chains []Chain, inputKeys []string, outputKeys []string)
 }
 
 func validateSeqChain(chain []Chain, inputKeys []string, outputKeys []string) error {
-	knownKeys := toSet(inputKeys)
+	knownKeys := util.ToSet(inputKeys)
 
 	for i, c := range chain {
 		// Check that chain has input keys that are in knownKeys
-		missingKeys := difference(c.GetInputKeys(), knownKeys)
+		missingKeys := util.Difference(c.GetInputKeys(), knownKeys)
 		if len(missingKeys) > 0 {
 			return fmt.Errorf(
 				"%w: chain at index %d is missing required input keys: %v",
@@ -45,7 +46,7 @@ func validateSeqChain(chain []Chain, inputKeys []string, outputKeys []string) er
 		}
 
 		// Check that chain does not have output keys that are already in knownKeys
-		overlappingKeys := intersection(c.GetOutputKeys(), knownKeys)
+		overlappingKeys := util.Intersection(c.GetOutputKeys(), knownKeys)
 		if len(overlappingKeys) > 0 {
 			return fmt.Errorf(
 				"%w: chain at index %d has output keys that already exist: %v",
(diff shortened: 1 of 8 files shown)


chains/sequential.go
```diff
@@ -0,0 +1,205 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/tmc/langchaingo/memory"
+	"github.com/tmc/langchaingo/schema"
+)
+
+// SequentialChain is a chain that runs multiple chains in sequence,
+// where the output of one chain is the input of the next.
+type SequentialChain struct {
```
FluffyKebab, line 14: Maybe export the memory field?

AMK9978:
There are changes in **chains_test.go**, **llm_test.go**, and other files that are not directly related to the changes for adding SequentialChain and SimpleSequentialChain. For the sake of the readability of our git track, isn't it better to have such changes in a different commit?

chains/sequential.go
tobiade: I planned to add memory support on another PR to keep this one scoped. There are some additional validations that need to occur when using memory e.g. validating there are no clashes with the memory keys and the input keys to the `SequentialChain`. 
It might be simple enough to add on this PR - can do that if you think it's worth doing.

tobiade:
The changes in `chains_test.go`,  `llm_test.go` and `map_reduce_test.go` are due to the uplifted `testLanguageModel` struct which I'm using for the `SimpleSequentialChain` and `SequentialChain` tests. I think those changes are small enough to keep on this PR. 

The alternative would be duplicating `testLanguageModel`, and then creating another PR to consolidate those changes. I'm not sure if that's worth the extra effort.

AMK9978:
@tobiade 
Why another PR? You can just commit those changes first, and then commit the other changes related to adding SequentialChain and SimpleSequential. Anyway, it's not that big deal.

tobiade:
Ah, I misunderstood. My mind jumped to different commit == different PR. That’s a fair suggestion. Agree - not a big deal, but happy to amend the commits if needed.

chains/sequential.go
FluffyKebab: Yeah, didn't think about the extra validations. This PR is already kind of large so you can add it a second if you think that is best.

chains/sequential.go
```diff
@@ -0,0 +1,205 @@
...
+		if _, ok := knownKeys[key]; !ok {
+			return fmt.Errorf("%w: output key %s is not in the known keys", ErrChainInitialization, key)
+		}
+	}
+
+	return nil
+}
+
+// toSet converts a list to a set.
+func toSet(list []string) map[string]struct{} {
```
AMK9978, line 73: What do u think about putting such utility functions inside a separate package named util or helper to be usable in other places? They seem to be general.

tobiade: My preference is to leave these functions close to where they are used until another use-case needs them. I do see your point though - maybe the maintainers can chime in here.

chains/sequential.go
```diff
@@ -0,0 +1,205 @@
...
+	return intersection
+}
+
+// Call runs the logic of the chains and returns the outputs. This method should
+// not be called directly. Use rather the Call, Run or Predict functions that
+// handles the memory and other aspects of the chain.
+func (c *SequentialChain) Call(ctx context.Context, inputs map[string]any, options ...ChainCallOption) (map[string]any, error) { //nolint:lll
+	for _, chain := range c.chains {
+		var err error
+		inputs, err = Call(ctx, chain, inputs, options...)
```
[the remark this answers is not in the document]
tobiade: Tbh I agree this could be clearer. I (blindly) did this to keep some level of parity with Python [source](https://github.com/hwchase17/langchain/blob/9cb2347453015d661520ec78d0f2f1e5968e0d92/langchain/chains/sequential.py#L176-L177):
```python
_input = chain.run(_input, callbacks=_run_manager.get_child(f"step_{i+1}"))
```

Is this clearer?
```go
var outputs map[string]any
var err error
for _, chain := range c.chains {
  outputs, err = Call(ctx, chain, inputs, options...)
  if err != nil {
    return nil, err
  }
  inputs = outputs
}
```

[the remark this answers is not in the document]
cduggn: @tobiade adding a comment on the original might be good enough

chains/sequential.go
tobiade: I thought about this again and moved the functions to a `util` package. Given the rate of change of this project, it's probably worth bubbling up a `util` package that others can see and can take advantage of if needed.

Thanks 👍🏾

chains/sequential.go
tobiade: I thought about this again and moved the functions to a `util` package. Given the rate of change of this project, it's probably worth bubbling up a `util` package that others can see and can take advantage of if needed.

Thanks 👍🏾

tobiade:
Hey @tmc - anything I need to do get this merged?

tobiade:
@tmc I've rebased this branch to pull in the latest changes 👍🏾

PR #184, opened 2023-07-10
openaillm: add support for openai organizations

lowczarc:
This PR adds the support for openai organzations.
Without this it's only possible to use the default organization of an account.

[Python Implementation](https://github.com/hwchase17/langchain/blob/master/langchain/llms/openai.py#L230)

As for the tests, I don't know if it's necessary since it's only adding a header to the openAI request and llms/openai doesn't seems to contain tests anyway

### PR Checklist

### DIFF
--- a/llms/openai/internal/openaiclient/openaiclient.go
+++ b/llms/openai/internal/openaiclient/openaiclient.go
@@ -11,10 +11,11 @@ var ErrEmptyResponse = errors.New("empty response")
 
 // Client is a client for the OpenAI API.
 type Client struct {
-	token      string
-	Model      string
-	baseURL    string
-	httpClient Doer
+	token        string
+	Model        string
+	baseURL      string
+	organization string
+	httpClient   Doer
 }
 
 // Option is an option for the OpenAI client.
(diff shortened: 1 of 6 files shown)


llms/openai/internal/openaiclient/openaiclient.go
```diff
@@ -11,10 +11,11 @@ var ErrEmptyResponse = errors.New("empty response")
 
 // Client is a client for the OpenAI API.
 type Client struct {
```
tobiade, line 13: Possibly out of scope for this PR, but worth considering: We could define a `transport` which is used by the `Client` that sets cross-cutting headers so we can avoid duplicating that logic across multiple files i.e. `embedding.go`, `chat.go`, etc

```go
// Implements http.RoundTripper
type transport struct {
    token string
    organization string
}

func (t *transport) RoundTrip(req *http.Request) (*http.Response, error) {
    req.Header.Set("Content-Type", "application/json")
    req.Header.Set("Authorization", "Bearer "+t.token)
    if t.organization != "" {
        req.Header.Set("OpenAI-Organization", t.organization)
    }
    return http.DefaultTransport.RoundTrip(req)
 } 
```

Alternatively, we could define a single `newHTTPRequest` function that sets all the headers as required.

tmc (approved):
LGTM

PR #199, opened 2023-07-17
prompts: FewShotPrompt implementation

### LINKED ISSUE #196: feat: FewShotPromptTemplate
It's a clean way to have `FewShotPromptTemplate`, to implement the ConstitutionalChain, and also enable people to build prompts in that way.
[source](https://github.com/hwchase17/langchain/blob/master/langchain/prompts/few_shot.py)

I'd like to work on it.

AMK9978:
### Description
Added FewShotPrompt. Fixes #196.
[Python source](https://github.com/hwchase17/langchain/blob/master/langchain/prompts/few_shot.py)

### PR Checklist

### DIFF
--- a/prompts/few_shot.go
+++ b/prompts/few_shot.go
@@ -1,0 +1,189 @@
+package prompts
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+)
+
+var (
+	// ErrNoExample is returned when none of the Examples and ExampleSelector are provided.
+	ErrNoExample = errors.New("no example is provided")
+	// ErrExamplesAndExampleSelectorProvided is returned when there are no Examples and ExampleSelector.
+	ErrExamplesAndExampleSelectorProvided = errors.New("only one of 'Examples' and 'example_selector' should be" +
+		" provided")
+)
+
+// FewShotPrompt contains fields for a few-shot prompt.
+type FewShotPrompt struct {
+	// Examples to format into the prompt. Either this or ExamplePrompt should be provided.
+	Examples []map[string]string
+	// ExampleSelector to choose the examples to format into the prompt. Either this or Examples should be provided.
+	ExampleSelector ExampleSelector
+	// ExamplePrompt is used to format an individual example.
+	ExamplePrompt PromptTemplate
+	// A prompt template string to put before the examples.
+	Prefix string
+	// A prompt template string to put after the examples.
+	Suffix string
+	// A list of the names of the variables the prompt template expects.
+	InputVariables map[string]any
+	// Represents a map of variable names to values or functions that return values. If the value is a function, it will
+	// be called when the prompt template is rendered.
+	PartialVariables map[string]any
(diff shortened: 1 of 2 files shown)


prompts/few_shot.go
```diff
@@ -0,0 +1,189 @@
...
+	// String separator used to join the prefix, the examples, and suffix.
+	ExampleSeparator string
+	// The format of the prompt template. Options are: 'f-string', 'jinja2'.
+	TemplateFormat TemplateFormat
+	// Whether to try validating the template.
+	ValidateTemplate bool
+}
+
+// FewShotCallOptions contains optional fields for FewShotPrompt.
+type FewShotCallOptions struct {
```
tmc, line 43: if we're putting multiple field sin here then the options argument to New shouldn't be variadic

AMK9978: I wanted to let developers not provide any option instead of passing an empty CallOption instance. I guess we have the same with the [llms.options](https://github.com/tmc/langchaingo/blob/main/llms/options.go).

AMK9978:
@FluffyKebab 
Could you take a look at this PR? It's essential for another PR which is ConstitutionalChain.

prompts/few_shot_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+// ExampleSelector is not included because ExampleSelector is not implemented.
+type testCaseStruct struct {
```
FluffyKebab, line 11: I feel like this could be defined in the test function instead.

prompts/few_shot_test.go
```diff
@@ -0,0 +1,156 @@
...
+			true,
+			"template: template:1:23: executing \"template\" at <.wrong_content>: map has no entry for key " +
+				"\"wrong_content\"",
+		},
+	}
+	for _, tc := range testCases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			runTest(t, tc)
```
FluffyKebab, line 112: Do we need a separate function for this?

prompts/few_shot.go
FluffyKebab: I don't think having options to the constructor is necessary for this.

AMK9978: So you mean removing both the variadic argument and FewShotCallOptions?

AMK9978: What about default values?


```
		ExampleSeparator: "\n\n",
		TemplateFormat:   TemplateFormatGoTemplate,
		ValidateTemplate: true,
```
these are their default values. I can set their type as a pointer to let developers set nil and I set their default. I want them to be able to use default values without explicitly setting them. Thoughts?

AMK9978:
@FluffyKebab @tmc 
I updated the code and removed FewShotCallOptions. I will appreciate it if you take another look at it.

Issue #125, opened 2023-06-14
Flushing ChatHistory based on token size


Struki84:
Hey everyone!

I'm currently working on building a test chatbot using langchain-go and I need to be able to flush messages from the chat history when the token size of the full prompt hits a certain limit.

To tackle this, I've been digging into the codebase and exploring options to contribute to the repo. The python version has this nifty reference to **`BaseLanguageModel`**, which handles the logic for measuring the token size of the stored memory buffer. However, the Go version doesn't yet have anything similar built-in, so I ended up using the [titoken-go module](https://github.com/pkoukk/tiktoken-go) to get the token size of my history.

Since I'm still a bit new to GoLang, I was wondering if this difference in design approach is something specific to Golang or just a design choice made by the author. As far as I understand, Golang has its own version of inheritance through embeddings so I created my own memory wrapper by embedding the memory `Buffer` struct. 

```go
package memory

import (
	"github.com/pkoukk/tiktoken-go"
	mem "github.com/tmc/langchaingo/memory"
	"github.com/tmc/langchaingo/schema"
)

type AsaiMemory struct {
	*mem.Buffer
	Encoding      string
	EncodingModel string
	TokenLimit    int
	Messages      []schema.ChatMessage
}

func NewAsaiMemory() *AsaiMemory {
	m := AsaiMemory{
		Buffer:        mem.NewBuffer(),
		Encoding:      "",
		EncodingModel: "gpt-3.5-turbo",
		TokenLimit:    2800,
	}

	return &m
}

func (m *AsaiMemory) LoadMemory() error {
	m.Messages = m.ChatHistory.Messages()
	return nil
}

func (m *AsaiMemory) GetMemoryString() string {
	bufferString, err := schema.GetBufferString(m.Messages, m.HumanPrefix, m.AIPrefix)
	if err != nil {
		return ""
	}
	return bufferString
}

func (m *AsaiMemory) TrimContext() error {
	tkm, err := tiktoken.EncodingForModel(m.EncodingModel)
	if err != nil {
		return err
	}

	bufferString, err := schema.GetBufferString(m.Messages, m.HumanPrefix, m.AIPrefix)
	if err != nil {
		return err
	}

	bufferLength := len(tkm.Encode(bufferString, nil, nil))

	if bufferLength > m.TokenLimit {
		for bufferLength > m.TokenLimit {
			m.Messages = m.Messages[1:]
[...]

PR #209, opened 2023-07-21
feat: Memory Integrations (chat history)

nidzola:
### PR Checklist

## Description
- adding the ability to port in any memory integration for the buffer
- by this logic, we can pass in any store that we would like to have to keep track of the chat history
Screenshot 2023-07-21 at 6 00 32 PM

### DIFF
--- a/memory/buffer.go
+++ b/memory/buffer.go
@@ -12,7 +12,7 @@ var ErrInvalidInputValues = errors.New("invalid input values")
 
 // Buffer is a simple form of memory that remembers previous conversational back and forths directly.
 type Buffer struct {
-	ChatHistory *ChatMessageHistory
+	ChatHistory schema.ChatMessageHistory
 
 	ReturnMessages bool
 	InputKey       string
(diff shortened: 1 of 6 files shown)


Struki84:
Very nice! As I understand this will allow me to implement a postgre based chat message history, which I attempted in https://github.com/tmc/langchaingo/pull/186

zivkovicn:
> Very nice! As I understand this will allow me to implement a postgre based chat message history, which I attempted in #186

exactly! 😄

Struki84:
@tmc Can we add `option.WithTokenLimit()` or something of the kind here? I know python lib has TokenBuffer class to flush to chat msgs that overflow the token size for a model. Not sure if we need an entirely new buffer for that, so we might consider it... cc @FluffyKebab 

related to https://github.com/tmc/langchaingo/issues/125

nidzola:
> @tmc Can we add `option.WithTokenLimit()` or something of the kind here? I know python lib has TokenBuffer class to flush to chat msgs that overflow the token size for a model. Not sure if we need an entirely new buffer for that, so we might consider it... cc @FluffyKebab
> 
> related to #125

I will see when I grab time, to add the token limit, like they have in the python lib, maybe we can do it as a separate PR.

tmc (approved):
Could you add or augment an example showing use?

nidzola:
> Could you add or augment an example showing use?

`TestBufferMemoryWithChatHistoryOption` is mimicking the store (that could be anything) and is an example of usage, do you think we need smth more?

nidzola:
> @tmc Can we add `option.WithTokenLimit()` or something of the kind here? I know python lib has TokenBuffer class to flush to chat msgs that overflow the token size for a model. Not sure if we need an entirely new buffer for that, so we might consider it... cc @FluffyKebab
> 
> related to #125

@tmc @Struki84 added the token limit logic https://github.com/tmc/langchaingo/pull/217

PR #222, opened 2023-07-25
memory: Feature conversational retrieval interface fix and rename

nidzola:
### PR Checklist

## Description
- this PR fixes the Conversational retrieval interface to accept `schema.Memory` instead of the `memory.Buffer`, by that we can in future pass in any memory that we need (for example Buffer, TokenBuffer or any other that we add).
- I've also renamed `Buffer` to be `ConversationBuffer` and `TokenBuffer` to be `ConversationTokenBuffer` so that we are closer to the langchain python implementation, so that it is easier on the eyes when we are comparing what is what.
- updated `ChatMessageHistory` schema to support errors, we need that when we port in psql, mongo or any other driven history, so that we don't swallow the errors when/if we have them

### DIFF
--- a/agents/conversational_test.go
+++ b/agents/conversational_test.go
@@ -26,7 +26,7 @@ func TestConversationalWithMemory(t *testing.T) {
 		llm,
 		[]tools.Tool{tools.Calculator{}},
 		ConversationalReactDescription,
-		WithMemory(memory.NewBuffer()),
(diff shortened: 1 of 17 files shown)


nidzola:
@tmc if you could check this out, would like to merge it, b/c it fixes the Conversational retrieval.

PR #229, opened 2023-07-31
locallms: Split arguments with spaces in & remove quote around prompt

lowczarc:
I attempted to use Llama with LangChainGo utilizing the "local" LLM, but encountered some issues when I tried to send arguments (specifically the model path) to the binary.

I faced an error when I tried using local.WithArgs("-m <path_to_the_model> -p"). It seemed as if the list of arguments was being sent to exec.CommandContext as a single argument.

Once I got it up and running, I noticed that extra quotes were added to the prompt.

To me, it seems like it was designed to be parsed by Bash, but it doesn't appear to be so.

I'm not entirely sure if this is actually a bug, or perhaps I used it incorrectly. If it's supposed to work like this, feel free to close this PR.

### PR Checklist

### DIFF
--- a/llms/local/internal/localclient/completions.go
+++ b/llms/local/internal/localclient/completions.go
@@ -3,7 +3,6 @@ package localclient
 import (
 	"context"
 	"os/exec"
-	"strconv"
 )
 
 type completionPayload struct {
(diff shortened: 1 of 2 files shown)


tmc:
This all needs a good refactor but this seems like a good step.

PR #234, opened 2023-08-05
bugfix: sql null column error

felixgunawan:
Currently when using sql chain on tables with nullable columns, it will return converting null to string error is not supported. To mitigate this issue, I'm suggesting to use sql.NullString to handle null values.
Fixes https://github.com/tmc/langchaingo/issues/235

### PR Checklist


zivkovicn (commented):
could you add the test for this?

felixgunawan:
done @zivkovicn , please help to review again

PR #247, opened 2023-08-11
vectorstore/options: Extending Vectorstore options to accept optional embedder

nidzola:
### PR Checklist

## Description
- this option allows us to use different embedders, if we have single vectorstore (like posgresql) which can handle (storing/using) multiple embeddings.

### Usage example:
```
....
        // dummy list of embedders
        embedders := map[string]embedder.Embedder{
		"openai": openaiEmbedder,
		"huggingface": huggingfaceEmbedder,
		"vertexai": vertexaiEmbedder,
		"someother": someotherEmbedder,
	}
        emd := embedders["openai"] // dummy example pulling the embedders from the list of supported embedders

 	result, err := chains.Run(
		context.TODO(),
		chains.NewRetrievalQAFromLLM(
			llm,
			vectorstores.ToRetriever(store, 5, vectorstores.WithNameSpace(id), vectorstore.WithEmbedder(emb)),
		),
		"What colors is each piece of furniture next to the desk?",
	)
....
```
from this point on, our `store` when the `AddDocuments` or `SimilaritySearch` methods are called from the vectorstore interface, we can actually use the needed embedder and not be tied to only one.

for example
```
// Store is a wrapper around the pinecone rest API and grpc client.
type Store struct {
	some store configuration where we could have or not have embedder
}

funct (s Store) AddDocuments(ctx context.Context, docs []schema.Document, options ...vectorstores.Option) error {
     opts := v.getOptions(options...)
     vectors, err := options.Embeder.EmbedDocuments(ctx, texts)
   ....
}
````

### DIFF
--- a/vectorstores/options.go
+++ b/vectorstores/options.go
@@ -1,5 +1,7 @@
 package vectorstores
 
+import "github.com/tmc/langchaingo/embeddings"
+
 // Option is a function that configures an Options.
 type Option func(*Options)
 
@@ -8,6 +10,7 @@ type Options struct {
 	NameSpace      string
 	ScoreThreshold float64
 	Filters        any
+	Embedder       embeddings.Embedder
 }
 
 // WithNameSpace returns an Option for setting the name space.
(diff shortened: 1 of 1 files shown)


nidzola:
@tmc

PR #257, opened 2023-08-20
callbacks: add standard interface and logger

FluffyKebab:
This adds an interface for callbacks and incorporates it into tools, the agent executor, the vector store retriever, all LLMs and the  LLM chain. Incorporating callbacks into all chains is yet do be done. 

### PR Checklist

### DIFF
--- a/callbacks/callbacks.go
+++ b/callbacks/callbacks.go
@@ -1,0 +1,19 @@
+package callbacks
+
+import (
+	"github.com/tmc/langchaingo/llms"
+	"github.com/tmc/langchaingo/schema"
+)
+
+type CallbackHandler interface {
+	HandleText(text string)
+	HandleLLMStart(prompts []string)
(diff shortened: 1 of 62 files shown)


callbacks/callbacks.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"github.com/tmc/langchaingo/llms"
+	"github.com/tmc/langchaingo/schema"
+)
+
+// Handler is the interface that allows for hooking into specific parts of an
+// LLM application.
+type Handler interface {
+	HandleText(text string)
```
tmc, line 11: I think we should propagate ctx in these calls, thoughts?

tmc (commented):
Great work, comment on the interface.

tmc:
I went ahead and rebased and threaded ctx through.

Issue #273, opened 2023-08-29
[Feature Request] add markdown header text splitter


edocevol:
Add new Markdown splitter, Markdown syntax have more information for Completion Context. If you have many source documents with HTML format, you can purify ([bluemonday](http://github.com/microcosm-cc/bluemonday)) and convert ([html-to-markdown](https://github.com/JohannesKaufmann/html-to-markdown)) to Markdown easily.
 
-------

For example, split the [CONTRIBUTING.md](https://github.com/tmc/langchaingo/blob/54eed7e75e47b9269e799562deb14def5618840a/CONTRIBUTING.md) and get the below chunks.

```
# Contributing to langchaingo
First off, thanks for taking the time to contribute! ❤️
All types of contributions are encouraged and valued. See the [Table of Contents](#table-of-contents) for different ways
to help and details about how this project handles them. Please make sure to read the relevant section before making
your contribution. It will make it a lot easier for us maintainers and smooth out the experience for all involved. The
community looks forward to your contributions. 🎉

---

# Contributing to langchaingo
> And if you like the project, but just don't have time to contribute, that's fine. There are other easy ways to support
> the project and show your appreciation, which we would also be very happy about:
> - Star the project
> - Tweet about it
> - Refer this project in your project's readme
> - Mention the project at local meetups and tell your friends/colleagues

---

## Table of Contents
- [Code of Conduct](#code-of-conduct)
- [I Have a Question](#i-have-a-question)

---

## Table of Contents
- [I Want To Contribute](#i-want-to-contribute)
  - [Reporting Bugs](#reporting-bugs)
    - [Before Submitting a Bug Report](#before-submitting-a-bug-report)
    - [How Do I Submit a Good Bug Report?](#how-do-i-submit-a-good-bug-report)
  - [Suggesting Enhancements](#suggesting-enhancements)
    - [Before Submitting an Enhancement](#before-submitting-an-enhancement)
    - [How Do I Submit a Good Enhancement Suggestion?](#how-do-i-submit-a-good-enhancement-suggestion)

---

## Table of Contents
  - [Your First Code Contribution](#your-first-code-contribution)
    - [Make Changes](#make-changes)
      - [Make changes in the UI](#make-changes-in-the-ui)
      - [Make changes locally](#make-changes-locally)
    - [Commit your update](#commit-your-update)
    - [Pull Request](#pull-request)
    - [Your PR is merged!](#your-pr-is-merged)

---

## Code of Conduct
[...]

PR #274, opened 2023-08-29
textsplitter: add markdown header text splitter

edocevol:
For issue: #273 

### PR Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -59,6 +59,10 @@ require (
 	github.com/shopspring/decimal v1.2.0 // indirect
 	github.com/spf13/cast v1.3.1 // indirect
 	github.com/temoto/robotstxt v1.1.2 // indirect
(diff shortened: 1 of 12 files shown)


sssvip:
Good job!

tmc:
Can you include rationale for this library vs Goldmark?

edocevol:
> Can you include rationale for this library vs Goldmark?

In my previous work experience, I used golang-commonmark, and I will check if Goldmark can be used as a replacement.

edocevol:
> Can you include rationale for this library vs Goldmark?

I have made some attempts. Goldmark uses a "tree" data structure to represent the nodes of Markdown, while golang-commonmark/markdown uses an array, and each node is delimited by Open and Close tags. Both libraries can be used, but migrating to Goldmark requires a lot of effort.

tmc (approved):
Feels like we should introduce an interface here and not take on the new direct dependency (we need to do more of that throughout)

Issue #278, opened 2023-08-31
[Feature Request] able to use chromadb as the vector stores


bjwswang:
We would like to use `chormadb(https://github.com/chroma-core/chroma)` as our vector store in langchaingo. Any plans on this feature ? If we do have a plan, I would like to work on this feature.

noodnik2:
FYI I'm working on a PR to add remote Chroma support (i.e., not in-process, but by accessing a remote Chromadb service).  If all goes well, expect an initial implementation for review later this week.

bjwswang:
> FYI I'm working on a PR to add remote Chroma support (i.e., not in-process, but by accessing a remote Chromadb service).  If all goes well, expect an initial implementation for review later this week.

well... i think we should contrubute to  chroma-go  before Introduce it into this project

bjwswang:
close this issue since this is merged

PR #282, opened 2023-09-06
vectorstores: Add remote chroma "vectorstore" adapter

noodnik2:
## implement [[Feature Request] able to use chromadb as the vector stores](https://github.com/tmc/langchaingo/issues/278) 

Proposed implementation of (remote) access to [Chroma](https://www.trychroma.com/) via [chroma-go](https://github.com/amikos-tech/chroma-go) library.  

### Eyes 👀 Wanted!

You can help this initiative by reviewing the proposed changes and leaving in-line PR comments!

### Scope / Success Criteria

The ambitions for the initial proposal are modest: create a "kernel" of support for Chrome that can be extended as new use cases or edge cases are discovered.

The initial test cases are based upon those already in place for the "Pinecone" adapter.  Additional tests are forthcoming for increased test coverage and new functionality.

### Motivations

My personal motivation for this stems from dual loyalties: 
- As an AI enthusiast / developer I've enjoyed the convenience and lack of "lock in" provided by Chroma from its "in-process" and "in-docker" access to localized vector storage for development workflows / experimentation; yet,
- As a Gopher, I've felt a little left out of the "LangChain" party, until now 😄  

### Due Diligence
Investigation into the various ways to interface LangChain directly with Chroma has led to the following conclusions:
- No "in-process" implementation yet exists for `golang`; therefore, initial access is being offered (only) to an (external) Chroma service.
- Discussions in [`langchaingo`'s Discord channel](https://discord.com/channels/1038097195422978059/1097608449266831390) and notes found in [chroma-core#1072](https://github.com/chroma-core/chroma/issues/1072) have suggested the choice to use the [chroma-go](https://github.com/amikos-tech/chroma-go) library's APIs to interact with the Chroma service.

### PR Checklist

- [X] Read the [Contributing documentation](https://github.com/tmc/langchaingo/blob/main/CONTRIBUTING.md).
- [X] Read the [Code of conduct documentation](https://github.com/tmc/langchaingo/blob/main/CODE_OF_CONDUCT.md).
- [X] Name your Pull Request title clearly, concisely, and prefixed with the name of the primarily affected package you changed according to [Good commit messages](https://go.dev/doc/contribute#commit_messages) (such as `memory: add interfaces for X, Y` or `util: add whizzbang helpers`).
- [X] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [X] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `Fixes #123`).
[...]

PR #283, opened 2023-09-07
feat: able to use chromadb as the vectorstore

bjwswang:
### PR Checklist

- [x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `Fixes #123`).   Fix: #278 


bjwswang:
Hi @noodnik2 Just let you know that I am quite open to collaborate. ^_^  This is just a alternative implementation by using chroma-go which you can take as a reference. During the conversation with the chroma-go owner, there are still lots of work in this client (see [issues]([noodnik2](https://github.com/noodnik2))) 
btw... the test data in similarity search is not correct... since I don't get a openai apikey...

noodnik2:
Hi @bjwswang, thanks for your reply.  Indeed, I think it'll be great to collaborate on this.  👍 

What do you think about continuing to work in parallel, and support each other through thoughtful feedback to each-other's PRs?  Both of our branches will improve if we take the best ideas from each other's work, which should coalesce towards a single solution which passes a superset of our test cases.  The end result then would reflect a combined effort no matter whose branch was eventually used.

From a brief look at your PR here for instance, I've become aware of some useful `chroma-go` library usage patterns I wasn't aware of before.  And, having seen some of your recent contributions to that library, I'm persuaded they're worth adopting.  😄 

This approach might also work well given our time-shift - it seems your night is my day, and vice-versa. 

I plan to take a closer look at this PR (and look for an updated branch if you've got one) later this afternoon and leave some in-line comments and/or questions.

Can you review [my Draft PR](https://github.com/tmc/langchaingo/pull/282) and leave some questions, comments or suggestions?

vectorstores/chromadb/chroma.go
```diff
@@ -0,0 +1,187 @@
...
+		{
+			URL:         fmt.Sprintf("%s://%s", s.scheme, s.host),
+			Description: "Chromadb server url for this store",
+		},
+	}
+	s.client = &chroma.Client{
+		ApiClient: chromaopenapi.NewAPIClient(configuration),
+	}
+
+	if _, err = s.client.Heartbeat(); err != nil {
```
noodnik2, line 64: At first, I suspected your calling the `Heartbeat` method here was new behavior, as I thought the other `New` functions (i.e., of the other `vectorstore` packages) didn't actually do any network I/O.  Reviewing them though I realized they _can_ initiate some (even if only asynchronously).

It might be a good idea to reflect in the comments/doc that this function actually makes a network call and will fail if immediately it can't reach the external service to make that clearer.

vectorstores/chromadb/chroma.go
```diff
@@ -0,0 +1,187 @@
...
+	nameSpace := s.getNameSpace(opts)
+
+	texts := make([]string, 0, len(docs))
+	ids := make([]string, len(docs))
+	for idx, doc := range docs {
+		texts = append(texts, doc.PageContent)
+		ids[idx] = fmt.Sprintf("%d", idx)
+	}
+
+	collection, err := s.client.CreateCollection(s.nameSpace, map[string]interface{}{}, true, s.embedder, s.distanceFunc)
```
noodnik2, line 82: I struggled with the idea of using the adapter's `NameSpace` concept to map directly to Chroma "Collections."  I resisted the temptation initially, almost gave in, then realized I believe even more strongly that they're separate things and that I believe we shouldn't "hide" Chroma collections behind the concept of `NameSpaces` in this API.

Please feel free to disagree, but consider my rationalizations:

1. Weaviate doesn't have "Namespaces" yet the idea is implemented using its other attributes for "filtering" objects (though I don't understand these well) such as "Class" and "Module", etc.   `NameSpace` seems to be implemented in that adapter using metadata filters, etc. which I felt could be done also in the Chroma adapter.
2. I've seen cases where Pinecone "NameSpaces" are composed - that is, a set of objects is defined to comprise a superset of _several_ namespaces.  That'd be hard to do in Chroma - e.g., search across several collections?

In any case, I'm very interested in hearing your feedback on this subject.

vectorstores/chromadb/chroma.go
```diff
@@ -0,0 +1,187 @@
...
+
+func (s Store) AddDocuments(ctx context.Context, docs []schema.Document, options ...vectorstores.Option) error {
+	opts := s.getOptions(options...)
+	nameSpace := s.getNameSpace(opts)
+
+	texts := make([]string, 0, len(docs))
+	ids := make([]string, len(docs))
+	for idx, doc := range docs {
+		texts = append(texts, doc.PageContent)
+		ids[idx] = fmt.Sprintf("%d", idx)
```
noodnik2, line 79: I find it humorous that you you assigned the `id` value to be the string of the index - basically the same thing as I did as a guess.

Is that the correct thing to do?  What is this value / used for?

vectorstores/chromadb/chroma.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	metadatas := make([]map[string]any, 0)
+	for i := 0; i < len(docs); i++ {
+		metadata := make(map[string]any)
+		for k, v := range docs[i].Metadata {
+			metadata[k] = v
+		}
+		metadata[s.textKey] = texts[i]
+		metadata[s.nameSpaceKey] = nameSpace
```
noodnik2, line 102: This is what I'm planning to do for `NameSpace` (see my lengthy comment above), and later use it as a filter when needed, as the Weaviate adapter does.

vectorstores/chromadb/chroma.go
```diff
@@ -0,0 +1,187 @@
...
+		return ErrEmbedderWrongNumberVectors
+	}
+
+	metadatas := make([]map[string]any, 0)
+	for i := 0; i < len(docs); i++ {
+		metadata := make(map[string]any)
+		for k, v := range docs[i].Metadata {
+			metadata[k] = v
+		}
+		metadata[s.textKey] = texts[i]
```
noodnik2, line 101: What is this (i.e., `textKey`) used for?  

I left it out of my branch since I don't know.  

The tests should fail if it's not needed, no??  😏

vectorstores/chromadb/chroma.go
```diff
@@ -0,0 +1,187 @@
...
+		return nil, err
+	}
+
+	result, err := collection.Query([]string{query}, int32(numDocuments), where, nil, s.includes)
+	if err != nil {
+		return nil, err
+	}
+
+	docs := make([]schema.Document, 0, len(result.Documents[0]))
+	for i := 0; i < len(result.Documents[0]); i++ {
```
noodnik2, line 140: I'll take a closer look at this, but I find it interesting that you're only using `result.Documents[0]`.  Is it true that there are never other elements returned in this list? 🤔

vectorstores/chromadb/embedingfunc.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+var _ chroma.EmbeddingFunction = wrappedEmbeddingFunction{}
+
+// wrappedEmbeddingFunction is a wrapper around an embeddings.
+type wrappedEmbeddingFunction struct {
+	embeddings.Embedder
+}
+
+func (embedder wrappedEmbeddingFunction) CreateEmbedding(documents []string) ([][]float32, error) {
```
noodnik2, line 17: One of my `TODO`s is to enable the injection of the embedder to use, so I'll see if this works for me.  However, I find it odd that you actually need to _run_ the embedder function and deal with its outputs.  I guess I'll find out more as I get to that...  Thanks for this example!

noodnik2 (commented):
Here's my first pass of comments on the changes you'd initially proposed.

I am curious to find out more about the changes you say are needed in `chroma-go` though, and am not sure if you're saying they're a "show-stopper" in terms of moving forward on integrating it with `langchaingo`. 

Anyways, I hope you'll take a few minutes to review [my proposed changes](https://github.com/tmc/langchaingo/pull/282/files) and give me some feedback: comments, questions or suggestions, and both of us can work towards the best implementation we can collectively come up with.

Regards,

-- Marty

bjwswang:
> Here's my first pass of comments on the changes you'd initially proposed.
> 
> I am curious to find out more about the changes you say are needed in `chroma-go` though, and am not sure if you're saying they're a "show-stopper" in terms of moving forward on integrating it with `langchaingo`.
> 
> Anyways, I hope you'll take a few minutes to review [my proposed changes](https://github.com/tmc/langchaingo/pull/282/files) and give me some feedback: comments, questions or suggestions, and both of us can work towards the best implementation we can collectively come up with.
> 
> Regards,
> 
> -- Marty

Hey! Sure. The principles inn this pr is "Maintain consistency with the other two vector database implementations". That explains most of the questions you menthioned.

vectorstores/chromadb/chroma.go
```diff
@@ -0,0 +1,187 @@
...
+var _ vectorstores.VectorStore = Store{}
+
+// New creates a new Store with options for chromadb.
+func New(opts ...Option) (vectorstores.VectorStore, error) {
+	s, err := applyClientOptions(opts...)
+	if err != nil {
+		return nil, err
+	}
+
+	configuration := chromaopenapi.NewConfiguration()
```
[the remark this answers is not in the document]
bjwswang: Actually I am fine with using a url or `basePath to keep same format with https://github.com/amikos-tech/chroma-go/blob/main/main.go#L61  which is actually my 1st version of this implementation.

btw... I am trying to convince/contribute some code to make this `NewCllient(basePath string)` https://github.com/amikos-tech/chroma-go/blob/main/main.go#L61  accept `NewClient(basePath,opts ...Option)` especially for `Logging`.. .you may have noticd that this client prints the responses for every api call.

vectorstores/chromadb/chroma.go
bjwswang: @noodnik2 seeing your comments to make me feel that we  do have struggled with the same things.... 
 
I changed many times on this `collection` or `namespace`... better to  request some comments from the langchain matainer.

vectorstores/chromadb/chroma.go
bjwswang: well...yes. seems that's how  chroma-go organizes the response

image

vectorstores/chromadb/embedingfunc.go
bjwswang: cause chroma-go use `float32` but  `float64` in langchaingo

vectorstores/chromadb/chroma.go
bjwswang: for these unclear defintions, we need a expert/maintainer later to explain why this is required in langchaingo. I don't see this in python version.

vectorstores/chromadb/chroma.go
bjwswang: These ids are options which is random string if empty in langchain python

vectorstores/chromadb/embedingfunc.go
bjwswang: This is one of my questions  needs to figure out with chroma-go owner.

Issue #306, opened 2023-10-04
[Urgent Feature Request] Return vector id from Pinecone


shrikarnag-q:
While using 

err = store.AddDocuments(context.Background(), []schema.Document{
        {
            PageContent: userData,
            Metadata: map[string]any{

            },
        },
    })
    
Please return pinecone vector id

Something like

id, err := store.AddDocuments(context.Background(), []schema.Document{
        {
            PageContent: userData,
            Metadata: map[string]any{

            },
        },
    })
    
    if err != nil {
		log.Fatal(err)
	}
	
This is very important to do update or delete vector from pinecone for which we can use other library. 
Help needed here urgently, stuck in a project.

tmc:
Sorry to hear that you're in a bit of a bind. Please feel free to open a PR suggesting the edit you'd like, in the mean time you can fork the code and use go.work files to replace your implementation.

PR #308, opened 2023-10-07
Adding Custom ID while Adding Document in Pinecone

shrikarnag-q:
### Adding Custom ID while pushing Document in Pinecone
While adding the document using 
`store.AddDocuments` method you can set the custom vector ID and upsert the document

Example:
`err = store.AddDocuments(context.Background(), []schema.Document{
		{
			PageContent: "Tokyo",
			Metadata: map[string]any{
				"population": 38,
				"area": 2190,
			},
		},
		vectorestores.WithID("_insertTheIDHere_"))`

Note: Incase the custom Id is not set, the method will generate UUID and assign as custom id

For the issue [#306](https://github.com/tmc/langchaingo/issues/306)


tmc:
Please update the PR description with a description of the change you are proposing when it is ready for review. In the mean time I'm going to make this draft - flip it back when the code is correct, tests are added and updated, and the PR description is updated to describe the change.

vectorstores/pinecone/grpcapi.go
```diff
@@ -47,7 +47,11 @@ func (s Store) grpcUpsert(
 	vectors [][]float32,
 	metadatas []map[string]any,
 	nameSpace string,
+	id string,
 ) error {
+	if id == "" {
+		id = uuid.New().String()
```
tmc, line 53: this behavior should be documented on an exported method.

vectorstores/options.go
```diff
@@ -44,3 +45,10 @@ func WithEmbedder(embedder embeddings.Embedder) Option {
 		o.Embedder = embedder
 	}
 }
+
+// WithID returns an Option for setting vectorId for document.
```
tmc, line 49: let's flesh this comment out some more.

shrikarnag-q (commented):
Have tried to correct the errors and issues. Please do check

vectorstores/options.go
```diff
@@ -44,3 +48,17 @@ func WithEmbedder(embedder embeddings.Embedder) Option {
...
+// This approach is beneficial when we need to perform deletions or updates on the vector within the Pinecone index.
+func WithID(id string) Option {
+	if id != "" {
+		return func(o *Options) {
+			o.ID = id
+		}
+	}
+	return func(o *Options) {
+		o.ID = uuid.NewString()
+	}
```
FluffyKebab, line 63: This logic should not be in the option, but in the AddDocuments method, because we need to create and UUID for each of the documents upserted if no IDs are specified and we dont know how many documents will be upserted in this function.

FluffyKebab (changes requested):

pieterclaerhout:
This should no longer be needed once https://github.com/tmc/langchaingo/pull/561 is merged…

tmc:
Closing those since we've moved on in those later PRs.

PR #318, opened 2023-10-13
vectorstore: openai apikey is must required only when embedder is nil

bjwswang:
### PR Checklist

### DIFF
--- a/examples/chroma-vectorstore-example/chroma_vectorstore_example.go
+++ b/examples/chroma-vectorstore-example/chroma_vectorstore_example.go
@@ -21,7 +21,6 @@ func main() {
 		chroma.WithChromaURL(os.Getenv("CHROMA_URL")),
 		chroma.WithOpenAiAPIKey(os.Getenv("OPENAI_API_KEY")),
 		chroma.WithDistanceFunction(chroma_go.COSINE),
-		chroma.WithCollectionName(uuid.New().String()),
 		chroma.WithNameSpace(uuid.New().String()),
 	)
 	if errNs != nil {
--- a/vectorstores/chroma/chroma.go
+++ b/vectorstores/chroma/chroma.go
@@ -30,11 +30,11 @@ type Store struct {
 	distanceFunction chromago.DistanceFunction
 	chromaURL        string
 	openaiAPIKey     string
-	collectionName   string
-	nameSpace        string
-	nameSpaceKey     string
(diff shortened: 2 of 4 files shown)


bjwswang:
@tmc @noodnik2  This pr contains a few changes:
1. make openai apikey must required only when embedder is nil.
2. collection name is duplicated with namespace. we keep namespace .
3.

tmc (commented):
I think this is a positive change -- Can you explain removing the collection name option?

bjwswang:
> I think this is a positive change -- Can you explain removing the collection name option

collection name is the concept in chroma which is the same thing as namespace

bjwswang:
> > I think this is a positive change -- Can you explain removing the collection name option
> 
> collection name is the concept in chroma which is the same thing as namespace

either we use collection or namespace

noodnik2:
> > I think this is a positive change -- Can you explain removing the collection name option
> 
> collection name is the concept in chroma which is the same thing as namespace

I originally kept namespace separate from collection because I thought it was useful.  Previously, all items in the same collection were kept separate between namespaces using a filter on the query.  That is - a "collection" is like a database which can store elements logically separated into separate namespaces, enforced through (filter) predicates. 

If that's not useful, then I have no concerns about removing support for this distinction.

noodnik2 (commented):
Apologies for the late review.  Work's been keeping me busy!

Consider removing the `getNamespacedFilter` and everything related to it.  Now that namespaces are just collections, there's no need for this filter, which previously was used to separate objects in the same collection but different namespaces. 
 
Presuming the ability to support namespaces within collections is not important, this change looks good.  I do suggest removing all vestiges of that support, now that it's not needed (describe in previous paragraph).

noodnik2:
> > I think this is a positive change -- Can you explain removing the collection name option
> 
> collection name is the concept in chroma which is the same thing as namespace

Perhaps I'm guilty of "overengineering" if in fact support for both together (i.e., multiple namespaces within a single collection) in my initial implementation was not needed or desired.  Good catch, if so!

noodnik2:
> btw... I don't think we should treat openai so different. Use a common embedder is good enough @noodnik2 any thought?

Not sure what you mean by "use a common embedder."  If I'm not mistaken, I had supported both the provided embedder, or the internal OpenAI embedder if one was not provided.

PR #319, opened 2023-10-14
bugfix: openai fix model option

Saerdna:
openai.New(openai.WithModel("model_name")) not work, it will use defaultEmbeddingModel as "text-embedding-ada-002"


tmc (approved):
Oi. This should have an assertion in a test.

Issue #307, opened 2023-10-05
Add Ollama support


tmc:
We should have Ollama support.

https://github.com/jmorganca/ollama/blob/main/docs/api.md

bjwswang:
@tmc This is so cool! Seems this is a pure go llm runner.

bjwswang:
@tmc I am interested in this . /assign @bjwswang

CyrilPeponnet:
I started toy around here https://github.com/tmc/langchaingo/pull/327

sausheong:
Exciting! Can't wait to use it!

tmc:
Closed via #327 ! Announced here: https://twitter.com/traviscline/status/1718400239441420391?s=61&t=mAFtj7oJZmLZnH-5yXho9w

PR #327, opened 2023-10-19
Adding ollama support

CyrilPeponnet:
### PR Checklist

Address #307


tmc:
Excited to see this!

CyrilPeponnet:
Example of use:
```go
	llm, err := ollama.New(
		ollama.WithModel("zephyr"),
	)
```

The model is required. One might still pass an alternative model with the WithModel llms.Options during a Call invocation. 

There is a bunch of options that you can set for the runner or the query itself that are not in the main llms.Options. 

Also as Ollama is templating the prompt query depending on the model you use, its disabled by default as it doesn't play well with agents and other templating done on the chain level.

CyrilPeponnet:
Adding the chat model support and all passing it back to draft for now.

CyrilPeponnet:
The chatModel support requires some knowledge about the the model that is running on the other side as the multiturn prompt is not handled by ollama by default. This implementation uses llama2 multi turn prompt syntax but you can provide your own if you are toying with Zephyr for instance.

I also added the embedding support even if the embedding provided by llamaccp is for now pretty limited, this might change in the future with the support of Bert models dedicated for document/query embedding.

tmc:
Fantastic work! sorry about the delay in review here. I'm actually thinking we should remove our "local" implementation as it was never really fleshed out and would love to lean further into ollama support!

embeddings/ollama/ollama.go
```diff
@@ -0,0 +1,73 @@
+package ollama
```
tmc, line 1: please include doc.go files in new packages with a package-level description.

embeddings/ollama/ollamachat/ollama_chat.go
```diff
@@ -0,0 +1,71 @@
...
+func NewChatOllama(opts ...ChatOption) (ChatOllama, error) {
+	o, err := applyChatClientOptions(opts...)
+	if err != nil {
+		return ChatOllama{}, err
+	}
+
+	return o, nil
+}
+
+func (e ChatOllama) EmbedDocuments(ctx context.Context, texts []string) ([][]float32, error) {
```
tmc, line 31: can you include a comment on this?

embeddings/ollama/ollamachat/ollama_chat.go
```diff
@@ -0,0 +1,71 @@
...
+			return nil, err
+		}
+
+		emb = append(emb, combined)
+	}
+
+	return emb, nil
+}
+
+func (e ChatOllama) EmbedQuery(ctx context.Context, text string) ([]float32, error) {
```
tmc, line 60: as above

tmc (approved):
There are some things to spruce up in here but overall this looks really great! Merging!

eliben:
@CyrilPeponnet can you please clarify why there are two very similar implementations of embeddings in `embeddings/ollama/ollama.go` and `embeddings/ollama/ollamachat/ollama_chat.go`? They seem essentially identical except that their `client` has a different type?

Issue #356, opened 2023-11-15
Different embedding results from EmbedQuery vs EmbedDocuments


eliben:
This sample (using a locally running Ollama):

```go
package main

import (
	"context"
	"fmt"
	"log"

	ollama_emb "github.com/tmc/langchaingo/embeddings/ollama"
	"github.com/tmc/langchaingo/llms/ollama"
)

func main() {
	ctx := context.Background()
	ollamaLLM, err := ollama.New(ollama.WithModel("llama2"))
	if err != nil {
		log.Fatal(err)
	}
	ollamaEmbeder, err := ollama_emb.NewOllama(ollama_emb.WithClient(*ollamaLLM))
	if err != nil {
		log.Fatal(err)
	}

	vec, err := ollamaEmbeder.EmbedQuery(ctx, "some text")
	if err != nil {
		panic(err)
	}

	fmt.Println("for single embedding:")
	fmt.Println(len(vec), "items")
	fmt.Println("first few:", vec[:16])

	texts := []string{
		"some text",
		"some other text",
		"nothing to see here",
		"some text",
	}
	vecs, err := ollamaEmbeder.EmbedDocuments(ctx, texts)
	if err != nil {
		panic(err)
	}

	fmt.Printf("got %d embeddings\n", len(vecs))
	for i, v := range vecs {
		fmt.Printf("[%d] len=%v, first few: %v\n", i, len(v), v[:16])
	}
}
```

I get the result:

```
for single embedding:
4096 items
first few: [-0.29370144 -0.9472216 0.15928172 2.9431767 2.1008253 0.42557585 0.4646868 -1.2639807 0.74785733 2.0149763 0.06937195 1.9570599 1.0287265 -0.59564227 0.4246622 0.55052483]
got 4 embeddings
[0] len=4096, first few: [-0.0023840654 -0.0076888897 0.001292939 0.023890676 0.017053049 0.0034545308 0.0037720064 -0.010260123 0.006070589 0.016356185 0.0005631136 0.015886059 0.008350491 -0.0048350124 0.0034471143 0.00446878]
[1] len=4096, first few: [0.00018647342 -0.0074219094 -0.0017224154 0.024820441 0.018676626 0.009781131 0.0017315162 -0.009649557 0.002851611 0.017244304 -0.003103958 0.01625366 0.0053226543 -0.0037721072 -0.0015953793 0.0027238138]
[...]

PR #358, opened 2023-11-19
embeddings: refactor EmbedderClient interface to reduce code duplication

eliben:
Starting with the vertex embedder, but this is applicable to others too.

The common code fixes the combined vector issue, similarly to #357, for #356

This replaces a completely duplicated method in two places with a common function; the same function can be used for the embedders of other LLMs as well, if this approach is deemed good.

### PR Checklist

### DIFF
--- a/embeddings/internal/embedderclient/embedderclient.go
+++ b/embeddings/internal/embedderclient/embedderclient.go
@@ -1,0 +1,45 @@
+package embedderclient
+
+import (
+	"context"
+
+	"github.com/tmc/langchaingo/embeddings"
+)
+
+// EmbedderClient is the interface LLM clients implement for embeddings.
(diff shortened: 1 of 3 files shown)


tmc (approved):
Much better, thank you. LGTM

eliben:
Thanks @tmc - I'd be happy to apply this to other embeddings backends, but it's not clear to me why there are two at all - why is there one instance of `LLM` and one of `Chat` with the same embeddings methods and implementations? E.g. https://github.com/tmc/langchaingo/blob/main/embeddings/ollama/ollamachat/ollama_chat.go#L31 and https://github.com/tmc/langchaingo/blob/main/embeddings/ollama/ollama.go#L32 -- could you clarify?

PR #362, opened 2023-11-21
embeddings: remove unnecessary mostly-duplicated implementations

eliben:
### PR Checklist


tmc (approved):
Thanks @eliben! Much cleaner.

eliben:
They're only removed from the `embeddings` directory.

Can you provide a [minimal, reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of what you're trying to accomplish?

eliben:
Sorry, I'm still having trouble understanding what you're trying to accomplish. The code that was removed duplicated all the embeddings implementation with the only difference being that it forwarded the `CreateEmbedding` method to the underlying client. This design didn't make a lot of sense, but it also wasn't clear what the goal is.

Can you provide an example of the actual problem you're trying to solve? Do you have a `llm.Chat` but not the underlying `llm.Client` and you're trying to get embeddings from it? How does this situation arise? Can you get embeddings from the client, not the chat? I'm confused about the duplication here.

eliben:
BTW, I recommend opening a new issue to discuss this - for more visibility. Feel free to @ me if you want to create one and provide an example

zivkovicn:
@eliben opened the issue, if you like you can find me on the discord as well to chat about this or even hop on the call if needed.

PR #365, opened 2023-11-22
examples/ernie-completion-examples: make this example a separate module

eliben:
Right now it's in the main module, the only example of this kind.
Align it with the other examples.
    
This is a two step change, and this is step 1; step 2 will be updating
this example's go.mod to rely on a newer version of langchaingo that
no longer provides this package.

This change has to comment out the example's contents for the lint
to pass -- the issue is that both the main module and this new module
provide the package.

### PR Checklist

### DIFF
--- a/examples/ernie-completion-example/ernie_completion_example.go
+++ b/examples/ernie-completion-example/ernie_completion_example.go
@@ -1,45 +1,35 @@
 package main
 
-import (
-	"context"
-	"fmt"
-	"log"
-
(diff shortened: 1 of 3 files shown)


tmc (approved):
Thanks, these should all definitely be separate modules, missed this.

eliben:
> Thanks, these should all definitely be separate modules, missed this.

Thanks for the quick turnaround; stay tuned for part 2 :)

PR #377, opened 2023-11-30
vectorstores: add pgvector

Abirdcfly:
### PR Checklist

### DIFF
--- a/vectorstores/pgvector/doc.go
+++ b/vectorstores/pgvector/doc.go
@@ -1,0 +1,3 @@
+// Package pgvector contains an implementation of the VectorStore
+// interface using pgvector.
+package pgvector
--- a/vectorstores/pgvector/pgvector.go
+++ b/vectorstores/pgvector/pgvector.go
@@ -1,0 +1,335 @@
+package pgvector
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v5"
+	"github.com/pgvector/pgvector-go"
+	"github.com/tmc/langchaingo/embeddings"
+	"github.com/tmc/langchaingo/schema"
+	"github.com/tmc/langchaingo/vectorstores"
+)
+
+const (
+	// pgLockIDEmbeddingTable is used for advisor lock to fix issue arising from concurrent
+	// creation of the embedding table.The same value represents the same lock.
+	pgLockIDEmbeddingTable = 1573678846307946494
+	// pgLockIDCollectionTable is used for advisor lock to fix issue arising from concurrent
+	// creation of the collection table.The same value represents the same lock.
+	pgLockIDCollectionTable = 1573678846307946495
+	// pgLockIDExtension is used for advisor lock to fix issue arising from concurrent creation
+	// of the vector extension. The value is deliberately set to the same as python langchain
+	// https://github.com/langchain-ai/langchain/blob/v0.0.340/libs/langchain/langchain/vectorstores/pgvector.py#L167
+	pgLockIDExtension = 1573678846307946496
+)
+
+var (
+	ErrEmbedderWrongNumberVectors = errors.New("number of vectors from embedder does not match number of documents")
+	ErrInvalidScoreThreshold      = errors.New("score threshold must be between 0 and 1")
+	ErrInvalidFilters             = errors.New("invalid filters")
+	ErrUnsupportedOptions         = errors.New("unsupported options")
+)
+
+// Store is a wrapper around the pgvector client.
+type Store struct {
+	embedder              embeddings.Embedder
+	conn                  *pgx.Conn
+	postgresConnectionURL string
+	embeddingTableName    string
+	collectionTableName   string
+	collectionName        string
+	collectionUUID        string
+	collectionMetadata    map[string]any
+	preDeleteCollection   bool
+}
+
+var _ vectorstores.VectorStore = Store{}
+
+// New creates a new Store with options.
+func New(ctx context.Context, opts ...Option) (Store, error) {
+	store, err := applyClientOptions(opts...)
+	if err != nil {
+		return Store{}, err
+	}
+	store.conn, err = pgx.Connect(ctx, store.postgresConnectionURL)
+	if err != nil {
+		return Store{}, err
+	}
+
+	if err = store.conn.Ping(ctx); err != nil {
+		return Store{}, err
+	}
+
+	if err = store.createVectorExtensionIfNotExists(ctx); err != nil {
+		return Store{}, err
+	}
+	if err = store.createCollectionTableIfNotExists(ctx); err != nil {
+		return Store{}, err
+	}
+	if err = store.createEmbeddingTableIfNotExists(ctx); err != nil {
+		return Store{}, err
+	}
+	if store.preDeleteCollection {
+		if err = store.RemoveCollection(ctx); err != nil {
+			return Store{}, err
+		}
+	}
+	if err = store.createOrGetCollection(ctx); err != nil {
+		return Store{}, err
+	}
+	return store, nil
+}
+
+func (s Store) createVectorExtensionIfNotExists(ctx context.Context) error {
+	tx, err := s.conn.Begin(ctx)
+	if err != nil {
+		return err
+	}
+	// inspired by
+	// https://github.com/langchain-ai/langchain/blob/v0.0.340/libs/langchain/langchain/vectorstores/pgvector.py#L167
+	// The advisor lock fixes issue arising from concurrent
+	// creation of the vector extension.
+	// https://github.com/langchain-ai/langchain/issues/12933
+	// For more information see:
+	// https://www.postgresql.org/docs/16/explicit-locking.html#ADVISORY-LOCKS
+	if _, err := tx.Exec(ctx, "SELECT pg_advisory_xact_lock($1)", pgLockIDExtension); err != nil {
+		return err
+	}
+	if _, err := tx.Exec(ctx, "CREATE EXTENSION IF NOT EXISTS vector"); err != nil {
+		return err
+	}
+	return tx.Commit(ctx)
+}
+
+func (s Store) createCollectionTableIfNotExists(ctx context.Context) error {
+	tx, err := s.conn.Begin(ctx)
+	if err != nil {
+		return err
+	}
+	// inspired by
+	// https://github.com/langchain-ai/langchain/blob/v0.0.340/libs/langchain/langchain/vectorstores/pgvector.py#L167
+	// The advisor lock fixes issue arising from concurrent
+	// creation of the vector extension.
+	// https://github.com/langchain-ai/langchain/issues/12933
+	// For more information see:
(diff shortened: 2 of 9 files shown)


vectorstores/pgvector/pgvector.go
```diff
@@ -0,0 +1,329 @@
...
+	// https://github.com/langchain-ai/langchain/blob/v0.0.340/libs/langchain/langchain/vectorstores/pgvector.py#L167
+	// The advisor lock fixes issue arising from concurrent
+	// creation of the vector extension.
+	// https://github.com/langchain-ai/langchain/issues/12933
+	// For more information see:
+	// https://www.postgresql.org/docs/16/explicit-locking.html#ADVISORY-LOCKS
+	if _, err = tx.Exec(ctx, "SELECT pg_advisory_xact_lock(1573678846307946494)"); err != nil {
+		return err
+	}
+	sql := fmt.Sprintf(`CREATE TABLE IF NOT EXISTS %s (
```
tmc, line 110: rather than string-assemble this can you use arguments to Exec (throughout) -- this protects against sql injection attacks.

vectorstores/pgvector/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package pgvector contains an implementation of the vectorStore
```
eliben, line 1: do you mean the `VectorStore` interface?

vectorstores/pgvector/pgvector.go
```diff
@@ -0,0 +1,329 @@
...
+		return err
+	}
+	// inspired by
+	// https://github.com/langchain-ai/langchain/blob/v0.0.340/libs/langchain/langchain/vectorstores/pgvector.py#L167
+	// The advisor lock fixes issue arising from concurrent
+	// creation of the vector extension.
+	// https://github.com/langchain-ai/langchain/issues/12933
+	// For more information see:
+	// https://www.postgresql.org/docs/16/explicit-locking.html#ADVISORY-LOCKS
+	if _, err := tx.Exec(ctx, "SELECT pg_advisory_xact_lock(1573678846307946495)"); err != nil {
```
eliben, line 86: Even if it's taken from elsewhere, it'd be good to describe the magic number passed to `pg_advisory_xact_lock`

vectorstores/pgvector/pgvector.go
```diff
@@ -0,0 +1,329 @@
...
+		return err
+	}
+	// inspired by
+	// https://github.com/langchain-ai/langchain/blob/v0.0.340/libs/langchain/langchain/vectorstores/pgvector.py#L167
+	// The advisor lock fixes issue arising from concurrent
+	// creation of the vector extension.
+	// https://github.com/langchain-ai/langchain/issues/12933
+	// For more information see:
+	// https://www.postgresql.org/docs/16/explicit-locking.html#ADVISORY-LOCKS
+	if _, err = tx.Exec(ctx, "SELECT pg_advisory_xact_lock(1573678846307946494)"); err != nil {
```
eliben, line 107: ditto

vectorstores/pgvector/pgvector.go
```diff
@@ -0,0 +1,329 @@
...
+PRIMARY KEY (uuid))`, s.embeddingTableName, s.collectionTableName)
+	if _, err = tx.Exec(ctx, sql); err != nil {
+		return err
+	}
+	return tx.Commit(ctx)
+}
+
+func (s Store) AddDocuments(ctx context.Context, docs []schema.Document, options ...vectorstores.Option) error {
+	opts := s.getOptions(options...)
+	if opts.Embedder != nil || opts.ScoreThreshold != 0 || opts.Filters != nil || opts.NameSpace != "" {
```
eliben, line 154: This is confusing to read, it's unhappy if `Embedder` is provided? Can you explain in a comment what's going on?

vectorstores/pgvector/pgvector.go
```diff
@@ -0,0 +1,329 @@
...
+
+func (s Store) createOrGetCollection(ctx context.Context) (string, error) {
+	sql := fmt.Sprintf(`INSERT INTO %s (uuid, name, cmetadata)
+		VALUES($1, $2, $3) ON CONFLICT DO NOTHING`, s.collectionTableName)
+	_, err := s.conn.Exec(ctx, sql, uuid.New().String(), s.collectionName, s.collectionMetadata)
+	if err != nil {
+		return "", err
+	}
+	sql = fmt.Sprintf(`SELECT uuid FROM %s WHERE name = $1 ORDER BY name limit 1`, s.collectionTableName)
+	rows, err := s.conn.Query(ctx, sql, s.collectionName)
```
eliben, line 284: Can you use `QueryRow` here (and elsewhere where you only expect to scan a single row)?

vectorstores/pgvector/pgvector.go
```diff
@@ -0,0 +1,329 @@
...
+	var collectionUUID string
+	for rows.Next() {
+		if err = rows.Scan(&collectionUUID); err != nil {
+			return "", err
+		}
+	}
+	return collectionUUID, nil
+}
+
+func (s Store) getOptions(options ...vectorstores.Option) vectorstores.Options {
```
eliben, line 298: Add a method comment explaining what this is doing

vectorstores/pgvector/pgvector.go
```diff
@@ -0,0 +1,329 @@
...
+	opts := vectorstores.Options{}
+	for _, opt := range options {
+		opt(&opts)
+	}
+	return opts
+}
+
+func (s Store) getNameSpace(opts vectorstores.Options) string {
+	if opts.NameSpace != "" {
```
eliben, line 307: I'm not sure I follow the logic here; if it's `""`, this returns `""`, so why the `if`?

vectorstores/pgvector/pgvector.go
```diff
@@ -0,0 +1,329 @@
...
+
+func (s Store) getScoreThreshold(opts vectorstores.Options) (float32, error) {
+	if opts.ScoreThreshold < 0 || opts.ScoreThreshold > 1 {
+		return 0, ErrInvalidScoreThreshold
+	}
+	return opts.ScoreThreshold, nil
+}
+
+// TODO: should support more types like {"key1": {"key2":"values2"}} or {"key": ["value1", "values2"]}.
+func (s Store) getFilters(opts vectorstores.Options) (map[string]any, error) {
```
eliben, line 321: Add a function comment

vectorstores/pgvector/pgvector_test.go
```diff
@@ -0,0 +1,409 @@
...
+		t.Skip("Must set PGVECTOR_CONNECTION_STRING to run test")
+	}
+
+	if openaiKey := os.Getenv("OPENAI_API_KEY"); openaiKey == "" {
+		t.Skip("OPENAI_API_KEY not set")
+	}
+}
+
+func getTestCollectionName() string {
+	return fmt.Sprintf("test-collection-%s", uuid.New().String())
```
eliben, line 34: Nit: naming this `makeNewCollectionName` will make it clearer that it generates a new random name every time it's called, rather than "getting" some pre-set name

vectorstores/pgvector/doc.go
Abirdcfly: change all `vectorStore` -> `VectorStore` in doc.go in commit 46b24e7e50be901081b57a8cdb87f640fe4c4c76

vectorstores/pgvector/pgvector.go
Abirdcfly: IMO, these magic number just not overlap is enough, from https://www.postgresql.org/docs/current/functions-admin.html#FUNCTIONS-ADVISORY-LOCKS.

I used 1573678846307946495, 1573678846307946494, 1573678846307946493 just to distinguish it from [1573678846307946496 in langchain](https://github.com/langchain-ai/langchain/blob/v0.0.340/libs/langchain/langchain/vectorstores/pgvector.py#L174).

vectorstores/pgvector/pgvector.go
Abirdcfly: This idea comes from langchain: `https://github.com/langchain-ai/langchain/blob/v0.0.340/libs/langchain/langchain/vectorstores/_pgvector_ data_models.py#L27`, IMO that there is only one UUID for the same name in the Collection table.

vectorstores/pgvector/pgvector_test.go
Abirdcfly: Good Point.

vectorstores/pgvector/pgvector.go
Abirdcfly: Update code in commit 1faf360fc0a5ec31fcdb2560396c1680d5216fbe

vectorstores/pgvector/pgvector.go
Abirdcfly: I previously thought that using the embedder already in Store would be sufficient, but it makes sense to use the embedder override in opts.

Update code in commit 1faf360fc0a5ec31fcdb2560396c1680d5216fbe

vectorstores/pgvector/pgvector.go
eliben: OK, but still it'd be better to have these as package-level constants, something like `const pgLockId<purpose> = ....`

vectorstores/pgvector/pgvector.go
eliben: I'm not sure I follow the connection.

This code queries to extract a single value from the table, correct? I see only one `collectionUUID` which gets overwritten in every iteration of `rows.Next`.

If you're only expecting a single row, `QueryRow` will be cleaner than `Query`

Abirdcfly: Update code in commit b4d839d7e5805f615dabb221e1b1c45adc093d26

vectorstores/pgvector/pgvector.go
eliben: I think Travis maybe  means that you shouldn't `Sprintf` into `package sql` calls like `Exec`, but rather use the `args` of each method, e.g. look at the example at https://pkg.go.dev/database/sql#example-Conn.ExecContext

`
result, err := conn.ExecContext(ctx, `UPDATE balances SET balance = balance + 10 WHERE user_id = ?;`, id)
`

See how `id` is passed into the SQL query?

vectorstores/pgvector/pgvector.go
```diff
@@ -14,6 +14,12 @@ import (
 	"github.com/tmc/langchaingo/vectorstores"
 )
 
+const (
```
eliben, line 17: Add a comment saying these numbers are arbitrary and should just be different from each other

vectorstores/pgvector/pgvector.go
tmc: Yes that's what I meant.

eliben (approved):
I think my review comments have been addressed

On to @tmc to approve and merge if he's happy with the PR

tmc (approved):
LGTM!

tmc:
> > Please add/include an example showing usage.
> 
> Should I refer to the `vectorstores/pgvector/pgvector_test.go` file in `examples` dir to write an example of storing vectors using openai llm and pgvector? 🤔

That would work, I'd love to see more ollama examples but you can you openai, up to you.

PR #378, opened 2023-11-30
examples: move go.work creation to root

### LINKED ISSUE #375: go.work file
I propose adding a `go.work` file at the root of the repository, to facilitate working on the examples with the latest in-repo version of the `langchaingo` packages.

In the discussion at https://github.com/golang/go/issues/53502 it seems generally acceptable to commit a `go.work` file in case of a monorepo (a single git repo with multiple related modules), since it allows easier local development and doesn't affect users of the module.

I see in `examples/Makefile` there's provision to create separate `go.work` files for each example dir, but this isn't necessary; a single `go.work` at the root should be sufficient.

eliben:
We only need a single go.work file in the reposotory, and `go work use -r .` can recursively add all sub-modules to it, without requiring bash iteration.

Moves the creation of go.work into the root Makefile go.work is still in .gitignore so it won't be committed

Fixes #375

### PR Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,3 +52,8 @@ clean-lint-cache:
 build-examples:
 	for example in $(shell find ./examples -mindepth 1 -maxdepth 1 -type d); do \
 		(cd $$example; echo Build $$example; go mod tidy; go build -o /dev/null) || exit 1; done
+
+.PHONY: add-go-work
+add-go-work:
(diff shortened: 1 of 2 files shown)


tmc (approved):
Very cool! I wasn’t aware of that capability.

Issue #379, opened 2023-11-30
embeddings: clean up code using new inferaces/type


eliben:
#374 introduced `NewEmbedder` which can be used to clean up much of the code in the `embeddings` package, following #372

This is a tracking issue for that work.

eliben:
#396 is retrofitting TEI into this pattern

tmc:
@eliben was TEI the last instance? are we good to close?

eliben:
Nope there's still `huggingface`, I think you and @zivkovicn worked on it (based on `git log`). It'd be nice to retrofit that into the new pattern as well.

eliben:
There are a few more details here: https://github.com/tmc/langchaingo/issues/388#issuecomment-1836818255

tmc:
> There are a few more details here: [#388 (comment)](https://github.com/tmc/langchaingo/issues/388#issuecomment-1836818255)

Ah thanks, I had missed that comment. Closing that issue as we have a tagged release but will leave this issue open.

eliben:
@tmc and @zivkovicn any chance one of you wants to take a look at the huggingface embeddings implementation to make it work with this refactoring so we can close the issue?

I have almost no experience working with Hugging Face

PR #374, opened 2023-11-30
embeddings: refactor embedding interfaces to enable creation from Chat and LLM

eliben:
For #372 

### PR Checklist

- [x] Name your Pull Request title clearly, concisely, and prefixed with the name of the primarily affected package you changed according to [Good commit messages](https://go.dev/doc/contribute#commit_messages) (such as `memory: add interfaces for X, Y` or `util: add whizzbang helpers`).
- [x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `Fixes #123`).
- [x] Describes the source of new concepts.
- [x] References existing implementations as appropriate.
- [x] Contains test coverage for new functions.
- [x] Passes all [`golangci-lint`](https://golangci-lint.run/) checks.

PR #385, opened 2023-12-01
embeddings: use NewEmbedder for OpenAI embeddings

eliben:
This is the first use of the refactoring made in #374 -- removing the embeddings/openai directory entirely, since it can be now implemented with NewEmbedder without duplicating code. The test was moved one dir down and works with the new setup.

This is a model that can be used to similarly refactor the other embeddings/* subdirs.

This is a large PR because it also updates example go.mods

For #379

### PR Checklist


eliben:
@tmc for your review

@zivkovicn FYI

tmc (approved):
👏 LGTM

nidzola:
@eliben nice work! Tnx

PR #398, opened 2023-12-05
workflow: fix publish-docs workflow error

eliben:
Fix errors like https://github.com/tmc/langchaingo/actions/runs/7094309081

### PR Checklist

- [x] Name your Pull Request title clearly, concisely, and prefixed with the name of the primarily affected package you changed according to [Good commit messages](https://go.dev/doc/contribute#commit_messages) (such as `memory: add interfaces for X, Y` or `util: add whizzbang helpers`).
- [x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `Fixes #123`).

PR #399, opened 2023-12-05
workflow: run CI on pushes to main branch, not other branches

eliben:
Right now, each PR runs CI twice - once because it's a PR, and another time because it comes from a push to a branch that created the PR. See for example https://github.com/tmc/langchaingo/pull/398, causing:

- https://github.com/tmc/langchaingo/actions/runs/7094374726
- https://github.com/tmc/langchaingo/actions/runs/7094367361

This change makes it so we only run CI once per PR, because there's no need to run it on a branch push.

Now CI runs on either a PR or a push to `main`


eliben:
@tmc PTAL - note that this PR runs each of the 3 checks defined in the workflow YAML only once and not twice like other PRs.

PR #438, opened 2023-12-19
feat: use WithCallback to set openai callback option

chyroc:


tmc (approved):
This should be covered by a test but this looks correct to me.

PR #439, opened 2023-12-19
pgvector: support postgres schema

Tecnologer:
Added support for Postgres Schemas. 

The table is named `"mySchema"."myTable"` instead `"mySchema.myTable"`.


tmc (approved):
LGTM

Issue #440, opened 2023-12-20
pgvector DropTables() fails because collectionTable dropped before embeddingTable


mehulmshah:
The pgvector vectorstore has a `DropTables(ctx)` function that drops the `collectionTableName` Table, and then drops the `embeddingTable` Table.

Running this function causes the error:
```ERROR: cannot drop table "Collections" because other objects depend on it (SQLSTATE 2BP01)```

In order to resolve this, you simply have to change the order of dropping so that the `embeddingTableName` Table is dropped first.

i.e. Simply change the function to this:
```
	if _, err := s.conn.Exec(ctx, fmt.Sprintf(`DROP TABLE IF EXISTS %s`, s.embeddingTableName)); err != nil {
		return err
	}
	if _, err := s.conn.Exec(ctx, fmt.Sprintf(`DROP TABLE IF EXISTS %s`, s.collectionTableName)); err != nil {
		return err
	}

	return nil
```

tmc:
Fixed in #441

PR #441, opened 2023-12-20
pgvector: Fix DropTable Function

mehulmshah:
## Overview
See issue #440. `DropTables()` returns an error:
```
ERROR: cannot drop table "Collections" because other objects depend on it (SQLSTATE 2BP01)
```
Switched the order to drop the `s.embeddingTableName` Table first, and then the `s.collectionTableName` Table.

### PR Checklist

### DIFF
--- a/vectorstores/pgvector/pgvector.go
+++ b/vectorstores/pgvector/pgvector.go
@@ -271,10 +271,10 @@ func (s Store) Close(ctx context.Context) error {
 }
 
 func (s Store) DropTables(ctx context.Context) error {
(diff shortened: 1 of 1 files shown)


tmc (approved):
Hmm this really should be covered by a test, but LGTM. Thanks for your contribution.

If you have the time could you add a test that exercises this?

mehulmshah:
> Hmm this really should be covered by a test, but LGTM. Thanks for your contribution.
> 
> If you have the time could you add a test that exercises this?

Agreed. Sure thing, I'll find some time this month to add a test case for this. Also, thoughts on adding more functionality around ability to delete embeddings / collections by ID or other field(besides the `preDelete` function)?

PR #442, opened 2023-12-20
Docs: fix comment typo

chyroc:
### PR Checklist

### DIFF
--- a/chains/llm.go
+++ b/chains/llm.go
@@ -28,7 +28,7 @@ var (
 	_ callbacks.HandlerHaver = &LLMChain{}
 )
 
-// NewLLMChain creates a new LLMChain with an llm and a prompt.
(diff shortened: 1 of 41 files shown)


chains/llm.go
```diff
@@ -28,7 +28,7 @@ var (
 	_ callbacks.HandlerHaver = &LLMChain{}
 )
 
-// NewLLMChain creates a new LLMChain with an llm and a prompt.
+// NewLLMChain creates a new LLMChain with a llm and a prompt.
```
tmc, line 31: are you sure about this? I think because "llm" (which should be capitalized) starts with an "el" sound I think that "an" is more appropriate.

tmc (commented):
Thanks so much! I added a small change on an vs a.

PR #448, opened 2023-12-21
vectorstores: Add ids to the return when adding documents in any vectorstore

### LINKED ISSUE #306: [Urgent Feature Request] Return vector id from Pinecone

jaylalakiya:
When documents are added using AddDocument method in any vectorstore, previously it was returning nothing while now it returns list of string ids.

Fixes #306 

### PR Checklist


tmc (approved):
can you follow-on and document the methods in VectorStore?

PR #454, opened 2023-12-25
vectorstores: fix weaviate not return score from SimilaritySearch

Abirdcfly:
### PR Checklist

### DIFF
--- a/vectorstores/weaviate/weaviate.go
+++ b/vectorstores/weaviate/weaviate.go
@@ -170,6 +170,7 @@ func (s Store) SimilaritySearch(
 	return s.parseDocumentsByGraphQLResponse(res)
 }
 
+//nolint:cyclop
 func (s Store) parseDocumentsByGraphQLResponse(res *models.GraphQLResponse) ([]schema.Document, error) {
 	if len(res.Errors) > 0 {
 		messages := make([]string, 0, len(res.Errors))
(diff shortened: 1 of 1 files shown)


vectorstores/weaviate/weaviate.go
```diff
@@ -197,10 +201,19 @@ func (s Store) parseDocumentsByGraphQLResponse(res *models.GraphQLResponse) ([]s
 		if !ok {
 			return nil, ErrMissingTextKey
 		}
+		additional, ok := itemMap["_additional"].(map[string]any)
```
tmc, line 204: what are the conditions/guarantees that this is present?

tmc: if it isn't always present it seems like we should return the document without a populated score vs return an error earlier, thoughts?

Abirdcfly: > what are the conditions/guarantees that this is present?
`certainty` in `_additional` is weaviate's built-in attributes:
https://weaviate.io/developers/weaviate/api/graphql/additional-properties

and we add this in https://github.com/tmc/langchaingo/blob/e98eb1703a8d377ef3efeaa68fb24b33ef34e726/vectorstores/weaviate/weaviate.go#L262

but I agress we can return the document without a populated score.

tmc (approved):
LGTM

PR #467, opened 2023-12-29
textsplitter: handle singular markdown tokens in markdown_splitter

corani:
The markdown splitter was not correctly handling singular markdown tokens
like Hr or Fence. These tokens do not have a corresponding close type,
which was causing issues in the indexOfCloseTag function.

To fix this, a check has been added to return the start index if the
closeType is nil, indicating a singular token. This prevents the function
from incorrectly trying to find a close tag for these tokens and running
past the end of the document.


tmc (commented):
Can you include a test case for this?

tmc (approved):
after looking more closely, I'm fine with this change. Let's add some more tests.

corani:
I found some more issues with the markdown splitting / rendering. I'll try to do some work on that this week and submit a PR to fix what I can find. I'm now getting why python langchain took the shortcut of just splitting by a handful of regex 😄 (not that I'm saying that's a better approach!)

Issue #465, opened 2023-12-29
Refactor LLM interfaces to prepare for multi-modal models


eliben:
(discussion started in #410 and email)

----

I started looking at the chat vs. non-chat interfaces, beginning with OpenAI. These days, all OpenAI API calls are supposed to go through the chat endpoint -- see https://platform.openai.com/docs/guides/text-generation (the old non-chat endpoints are just for legacy models).

We can remove duplication stemming from these interfaces having overlapping but incompatible signatures:

```go
// LLM is a langchaingo Large Language Model.
type LLM interface {
    Call(ctx context.Context, prompt string, options ...CallOption) (string, error)
    Generate(ctx context.Context, prompts []string, options ...CallOption) ([]*Generation, error)
}

// ChatLLM is a langchaingo LLM that can be used for chatting.
type ChatLLM interface {
    Call(ctx context.Context, messages []schema.ChatMessage, options ...CallOption) (*schema.AIChatMessage, error)
    Generate(ctx context.Context, messages [][]schema.ChatMessage, options ...CallOption) ([]*Generation, error)
}
```

As it is right now, a single type can't implement both, so everything is duplicated. Also, the LLM interface is unsuitable for multi-modals because it takes a string. This may require quite a bit of breaking interfaces, unfortunately.

eliben:
There's another interface that seems related, with a somewhat similar signature:

```go
// LanguageModel is the interface all language models must implement.
type LanguageModel interface {
	// Take in a list of prompt values and return an LLMResult.
	GeneratePrompt(ctx context.Context, prompts []schema.PromptValue, options ...CallOption) (LLMResult, error)
	// Get the number of tokens present in the text.
	GetNumTokens(text string) int
}
```

Unless I'm mistaken, it seems to only be used in in `chains`.

eliben:
The idea of the main refactoring is to create a new interface for multimodal LLMs, with a single method, name TBD (could be `Complete` or `Call` or `Generate`, etc). The name of the interface is TBD as well (could be `LLM`, or `MultiModal` or maybe just `Model`?)

Looking at some of the backends that will end up being the most popular:

1. OpenAI: [the API](https://platform.openai.com/docs/guides/text-generation/chat-completions-api) takes an array of messages, each one with a role and an option type that can specify text an image (as a URL).
[...]

PR #468, opened 2023-12-29
Refactor LanguageModel interface away

eliben:
Use the LLM interface instead.

re #465


eliben:
@tmc PTAL, first step in the #465 refactoring. More red than green :+1: 

After you LGTM, please let me merge this PR. I recommend merging the commits as-is without squashing - this will make rollbacks easily if something breaks. I made the commits logical units.

tmc (approved):
Great refactor! Thanks @eliben!

PR #477, opened 2024-01-02
add run id by context

h3l:
### PR Checklist


eliben (changes requested):
You've marked: " Provide a description in this PR that addresses what the PR is solving, or reference the issue that it solves" in PR checklist, but I see no evidence of this in the PR description. What problem is this PR solving?

h3l:
@eliben Thank you for the reminder; it has been corrected.

chains/chains.go
```diff
@@ -44,6 +45,7 @@ func Call(ctx context.Context, c Chain, inputValues map[string]any, options ...C
 
 	callbacksHandler := getChainCallbackHandler(c)
 	if callbacksHandler != nil {
+		ctx = util.GenNewSubCtx(ctx)
```
tmc, line 48: I would expect something like NewContext/FromContext, but more broadly I think we need to design our approach to Tracing/OpenTelemetry as I think what you're ultimately after here is a Trace ID.

h3l: yes, i borrow `run id` and `parent run id` from langchain, but fundamentally speaking, we need something like OpenTelemetry to implement run id. 
A more fundamental question to consider is whether LangChiango should maintain alignment with LangChain.

tmc:
@h3l could you rebase this?

tmc:
This has gotten pretty stale -- @h3l are you open to rebasing this work or should I close it out?

PR #480, opened 2024-01-02
GenerateContent - initial experimental API with OpenAI implementation

eliben:
First sketch for #465, including an OpenAI implementation.

Also adding real tests that require API key to be set.

I'm marking this API as experimental so we can work on it without disrupting other code, initially.

### DIFF
--- a/llms/generatecontent.go
+++ b/llms/generatecontent.go
@@ -47,10 +47,13 @@ type BinaryContent struct {
 
 func (BinaryContent) isPart() {}
 
+// ContentResponse is the response returned by a GenerateContent call.
+// It can potentially return multiple response choices.
 type ContentResponse struct {
 	Choices []*ContentChoice
 }
 
+// ContentChoice is one of the response choices returned by GenerateModel calls.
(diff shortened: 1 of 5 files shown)


llms/generatecontent.go
```diff
@@ -0,0 +1,58 @@
...
+	Data     []byte
+}
+
+func (BinaryContent) isPart() {}
+
+type ContentResponse struct {
+	Choices []*ContentChoice
+}
+
+type ContentChoice struct {
```
tmc, line 54: Can you document this type?

llms/generatecontent.go
```diff
@@ -0,0 +1,58 @@
...
+
+// BinaryContent is content holding some binary data with a MIME type.
+type BinaryContent struct {
+	MIMEType string
+	Data     []byte
+}
+
+func (BinaryContent) isPart() {}
+
+type ContentResponse struct {
```
tmc, line 50: Can you document this type?

tmc (approved):
LGTM

PR #489, opened 2024-01-04
Clean up GeneratePrompt -- unused code, and simplify caller

eliben:
`GeneratePrompt` is an undocumented top-level function in llms - it's unused except in one place which has to work hard to adapt to it instead of just using `llm.Call` more simply -- update the single caller and simplify the code

`GenerateChatPrompt` is completely unused, so remove it.

re #465


tmc (approved):
Thanks for the cleanup. LGTM.

Issue #410, opened 2023-12-13
llms: Add Gemini Pro support + example


tmc:
With Gemini pro now available, we should have integration + an example.

eliben:
Happy to work on this within the next couple of weeks.

One interesting challenge is that Gemini are multimodal - they accept images as well as text as input

tmc:
Yes, I have the start of image support in https://github.com/tmc/langchaingo/pull/361 that I'd like to extend to support ollama's multi-modal support (in addition to Gemini)

eliben:
A key issue is that the `llms.LLM` interface has only `[]string` for its prompts, and a lot of code depends on it. We need to either break it and rewrite existing code, or add another interface.

tmc:
The chat interfaces supply a slightly richer surface area that we can extend but yes I agree we need a more flexible interface here. As we're pre-1.0 I'm open to considering a breaking change, it's pretty clear that text-only prompting is isn't sufficient.

eliben:
I still want to spend some time studying the chat interfaces; they could be ripe for a significant refactoring. Chats shouldn't have different interfaces from LLMs - they should just build on top. A chat should just wrap an LLM with some history/context. Haven't found time for this yet, though - but hopefully soon.

tmc:
Certainly open to collaborating on arriving at a good design!

xavidop:
happy to help here! I am GDE in ML

eliben:
Just a note that #465 was opened to specifically discuss the new interfaces needed to support multi-modal input. We'll design something that works for both OpenAI and Gemini and hopefully other models as well.

eliben:
Once #465 is done, adding a Google Gemini backend should be straightforward. I think the following plan makes sense:

1. Finish #465 
2. Add a new Google generative AI backend: this can either use the Google AI SDK (using an API key) or the Vertex SDK (using a GCP account) in the same backend
3. Once (3) works, remove the current `vertexai` backend since it's tuned for the old text-only PaLM model

xavidop:
@eliben I would suggest to use VertexAI SDK since the Google AI SDK is not available in all the countries yet

eliben:
@xavidop step (2) mentions that both can be supported via this interface. The user should be able to choose which to use when a client is created.

Would you like to help with this once the initial new interface from (1) is in place?

mrothroc:
[...]

PR #492, opened 2024-01-04
vertexai: add gemini support

mrothroc:
Fixes [#410](https://github.com/tmc/langchaingo/issues/410).

Adds gemini support to the vertexai client.  It will still use the old client by default.  If you want to use gemini, set the model name either on llm creation or when you make a call.  It supports dynamically choosing a model name.

It also uses the old client for embeddings, which don't appear to be supported in the new client just yet.


eliben:
Thanks, I will take a look. I was planning to tackle this after https://github.com/tmc/langchaingo/issues/465 because that will remove a lot of the duplicated interfaces and code and make stuff easier to add, but if you don't mind the extra work I'm happy to review this now and make adjustments later as #465 moves along

mrothroc:
I would welcome some rationalization of the interface.  The client code is structurally more complex than I'd like.  I did the best I could but the Google APIs are pretty different and stapling them together makes it a little messy.

It would be especially great to make a clean break and just support the genai client with the new interface from #465.  In that case, we could use what I did in this PR as a bridge to allow people to migrate over time.

eliben:
> I would welcome some rationalization of the interface. The client code is structurally more complex than I'd like. I did the best I could but the Google APIs are pretty different and stapling them together makes it a little messy.
> 

Indeed. I recommend you read the messages in #465 to see my plan; more details below

> It would be especially great to make a clean break and just support the genai client with the new interface from #465. In that case, we could use what I did in this PR as a bridge to allow people to migrate over time.

I want to reduce the core LLM interface to a single method, which is already in `llms.go` (added as part of #465):

```go
type Model interface {
	// GenerateContent asks the model to generate content from parts.
	GenerateContent(ctx context.Context, parts []ContentPart, options ...CallOption) (*ContentResponse, error)
}
```

And build all chat functionality on top of this; this should remove A LOT of code duplication because right now each provider has its own chat functionality copy-pasted from OpenAI.

My next step was to add a new provider called "googleai", that would ultimately let people use either https://pkg.go.dev/github.com/google/generative-ai-go@v0.5.0/genai or https://pkg.go.dev/cloud.google.com/go/vertexai@v0.6.0/genai transparently -- these are almost identical SDKs, and we can have an identical interface through langchaingo

I would only implement the new `Model` interface for this though, and then later when the chat is added my hope would be to remove the old Vertex PaLM provider entirely (except reusing it for embeddings for now, until the new vertex SDK supports this).

Please LMK if this makes sense - happy to elaborate

mrothroc:
Makes sense.  Plus one.  I could implement the Model interface for genai.  It would be a pretty easy based on what I've already done.

eliben:
> Makes sense. Plus one. I could implement the Model interface for genai. It would be a pretty easy based on what I've already done.

Yeah, I think that could work. I would recommend a new provider directory, something like `googleai` alongside the other directories in llms, so it would be `llmx/googleai`. You can start with just using the Google AI SDK (https://pkg.go.dev/github.com/google/generative-ai-go) implementing `Model` and we can take it from there. WDYT?

mrothroc:
Sounds good.  From what I see, this approach is almost identical to the Google library, so it should quite easy to implement this.  It is the end of my day now, but I can start on it tomorrow.

mrothroc:
I submitted a PR.  The genai API is still a little rough.  I couldn't get it to work with ADC.  Also, it doesn't appear to support images.  Anyway, it is a start for the basic use cases.

eliben (changes requested):
Leaving this just as a note that work moved to #497 which is a starting point from which to expand on gemini support in langchaingo

This PR can be closed or remain for reference, but shouldn't be merged

mrothroc:
Well, as I mentioned in the other comment, lack of support for ADC/services accounts/workload identity is a showstopper.  Without this PR, people (rather ironically) won't be able to use gemini-pro in Google Cloud Run or GKE.

eliben:
> Well, as I mentioned in the other comment, lack of support for ADC/services accounts/workload identity is a showstopper. Without this PR, people (rather ironically) won't be able to use gemini-pro in Google Cloud Run or GKE.

For the GCP side of things, you need the other SDK here: https://pkg.go.dev/cloud.google.com/go/vertexai/genai

The two should be mostly compatible, so it should be possible to use either through the langchaingo API once we hook things up, with client selection (maybe as an option, or just a separate "new client" function)

eliben:
Closing this since both vertex and googleai have Gemini models now. See #410

Issue #405, opened 2023-12-13
ollama: Use new chat endpoint


tmc:
As of https://github.com/jmorganca/ollama/releases/tag/v0.1.14 Ollama has a chat endpoint! we should use it!

PR #495, opened 2024-01-05
Add Ollama test and remove custom prompt for chat

eliben:
Following #405, the custom prompting is no longer needed since we now use the `/chat` endpoint of Ollama for chats.

This helps clean out a bunch of code!

Adding test for ollama that can be run locally by setting the `OLLAMA_TEST_MODEL` env var (otherwise skip).


tmc (approved):
LGTM

PR #497, opened 2024-01-05
feature: add new gemini llm supporting new experimental interface

mrothroc:
This adds a new "googleai" llm that supports the new Model interface.


llms/googleai/googleai_llm.go
```diff
@@ -0,0 +1,140 @@
...
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/google/generative-ai-go/genai"
```
eliben, line 11: Note that this is the SDK for ai.google.dev, which doesn't use GCP auth but rather just an API key like OpenAI - you can get an API key from that website

llms/googleai/googleai_llm_test.go
```diff
@@ -0,0 +1,89 @@
...
+			substr:  "",
+		},
+		{
+			name:    "SinglePart",
+			parts:   []llms.ContentPart{llms.TextContent{Text: "Say the word Panama and nothing else."}},
+			opts:    []llms.CallOption{},
+			wantErr: false,
+			substr:  "Panama",
+		},
+		// Following is disabled because the API throws this error:
```
eliben, line 44: You need `gemini-pro-vision` for image support

eliben: See https://ai.google.dev/tutorials/go_quickstart

llms/googleai/googleai_llm.go
```diff
@@ -0,0 +1,140 @@
...
+	var out genai.Part
+	var err error
+
+	switch p := part.(type) {
+	case llms.TextContent:
+		out = genai.Text(p.Text)
+	case llms.BinaryContent:
+		out = genai.Blob{MIMEType: p.MIMEType, Data: p.Data}
+	case llms.ImageURLContent:
+		out, err = downloadBytes(p.URL)
```
eliben, line 111: You'll probably want `genai.ImageData` here

llms/googleai/googleai_llm.go
```diff
@@ -0,0 +1,140 @@
...
+		}
+	}
+
+	metadata := make(map[string]any)
+	metadata[CITATIONS] = candidate.CitationMetadata
+	metadata[SAFETY] = candidate.SafetyRatings
+
+	return &llms.ContentChoice{
+		Content:        buf.String(),
+		StopReason:     fmt.Sprintf("%v", candidate.FinishReason),
```
eliben, line 137: Just use `.String()` instead of `Sprintf`

llms/googleai/googleai_llm_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"github.com/joho/godotenv"
+	"github.com/stretchr/testify/require"
+	"github.com/tmc/langchaingo/llms"
+	"google.golang.org/api/option"
+)
+
+func TestMain(m *testing.M) {
+	// This loads the API key if it is a .env file
+	_ = godotenv.Load()
```
eliben, line 16: We don't use `godotenv` elsewhere - let's just expect the env var to be set, like other tests do. The test should be skipped if the env var isn't set

Follow the example of `llms/openai/multicontent_test.go`

eliben (commented):
This is a good start, thanks! Here's a first round of comments

A general note: please add some documentation in comments, for conversion functions and other stuff too

mrothroc:
I replaced my tests with a copy of the one from openai and made the other adjustments as requested.  The one rather large issue here is the lack of support for ADC.  All of my code runs in either Cloud Run or GKE, and the best practice is to use service accounts with Workload Identity.  See here:
https://cloud.google.com/iam/docs/best-practices-service-accounts

I can't inject API Keys because that bypasses our audits and security process, which means I can't use this code.  Hope someone else finds it helpful.

eliben:
Thanks, this is a good start. 

> The one rather large issue here is the lack of support for ADC. All of my code runs in either Cloud Run or GKE, and the best practice is to use service accounts with Workload Identity. See here: https://cloud.google.com/iam/docs/best-practices-service-accounts
> 
> I can't inject API Keys because that bypasses our audits and security process, which means I can't use this code. Hope someone else finds it helpful.

See https://github.com/tmc/langchaingo/pull/492#issuecomment-1879434760

PR #498, opened 2024-01-06
googleai: set up options and embeddings

eliben:
Set up provider-specific options instead of exposing google client options to users

Add implementation of embedding + test

re #410

### DIFF
--- a/llms/googleai/googleai_llm.go
+++ b/llms/googleai/googleai_llm.go
@@ -167,5 +167,17 @@ func convertCandidate(candidate *genai.Candidate) (*llms.ContentChoice, error) {
 }
 
 // CreateEmbedding creates embeddings from texts.
-//func (g *GoogleAI) CreateEmbedding(ctx context.Context, texts []string) ([][]float32, error) {
-//}
+func (g *GoogleAI) CreateEmbedding(ctx context.Context, texts []string) ([][]float32, error) {
+	em := g.client.EmbeddingModel(g.opts.defaultEmbeddingModel)
+
+	var results [][]float32
+	for _, t := range texts {
+		res, err := em.EmbedContent(ctx, genai.Text(t))
+		if err != nil {
+			return results, err
+		}
+		results = append(results, res.Embedding.Values)
+	}
+
+	return results, nil
+}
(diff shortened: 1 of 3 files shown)


eliben:
@mrothroc please review

Note that I also renamed `New` to `NewGoogleAI` to prepare for a future where we'll have another constructor for creating a Vertex genai client that uses GCP auth but provides the same functionality

llms/googleai/googleai_llm.go
```diff
@@ -31,11 +32,18 @@ const (
...
+	for _, opt := range opts {
+		opt(&clientOptions)
+	}
+
+	gi := &GoogleAI{
+		opts: clientOptions,
+	}
 
-	client, err := genai.NewClient(ctx, options...)
+	client, err := genai.NewClient(ctx, option.WithAPIKey(clientOptions.apiKey))
```
tmc, line 46: I know there was some discussion regarding ADC on the issue -- does this preclude falling back to ADC?

tmc (approved):
LGTM

llms/googleai/googleai_llm.go
eliben: Nope, that will be a separate client hidden behind the abstraction in langchaingo -- see https://github.com/tmc/langchaingo/issues/410#issuecomment-1879717569

llms/googleai/googleai_llm.go
```diff
@@ -31,11 +32,18 @@ const (
 	SAFETY    = "safety"
 )
 
-// New creates a new GoogleAI struct.
-func New(ctx context.Context, options ...option.ClientOption) (*GoogleAI, error) {
-	gi := &GoogleAI{}
+// NewGoogleAI creates a new GoogleAI struct.
+func NewGoogleAI(ctx context.Context, opts ...Option) (*GoogleAI, error) {
```
tmc, line 36: Should this just be New to match the other packages?

tmc: From the discussion on the issue I'm guessing this is making room for the next step, so LGTM. Sorry about the delay here! I got bit by a bad fluish bug (covid test is negative though) and have had to rest a lot the last couple of days. Thanks!

eliben: The plan is to add something like `NewVertex` that will create the same `*GoogleAI` provider but using Vertex / GCP credentials, so users will select which one they want by invoking the appropriate constructor

PR #499, opened 2024-01-08
vectorstores: pgvector add option WithConn

Abirdcfly:
### PR Checklist


tmc (commented):
looking good but please address lint issues

vectorstores/pgvector/pgvector.go
```diff
@@ -55,33 +55,41 @@ func New(ctx context.Context, opts ...Option) (Store, error) {
 	if err != nil {
 		return Store{}, err
 	}
-	store.conn, err = pgx.Connect(ctx, store.postgresConnectionURL)
```
tmc, line 58: This constructor should return a pointer to Store

Abirdcfly: Please correct me if I am mistaken, but are you saying that the `func New(ctx context.Context, opts ...Option) (Store, error)` on line 53 should return a `*Store`?

tmc: Yes, (*Store, error), but we can correct that in another PR

PR #509, opened 2024-01-10
Link to LangChain Python (deserves credit)

carsonkahn-external:


eliben (approved):
This looks like a reasonable attribution to add

PR #514, opened 2024-01-16
Add callback handler method for GenerateContent start and end

eliben:
Re #465

PR #521, opened 2024-01-17
Finalize the majority of the model/LLM interface refactoring

eliben:
This is the big change for #465 building upon all the other work so far.

It's all a single PR because I wanted to test everything together; but it's broken into a sequence of logical commits that can be seen in isolation and reverted in isolation, if needed.

It's dependent on #514


tmc (approved):
Great work! LGTM

PR #525, opened 2024-01-17
Deprecate the LLM interface

### LINKED ISSUE #465: Refactor LLM interfaces to prepare for multi-modal models

eliben:
It's just an alias for Model now

Deprecating it means lint will complain when someone uses it directly. 

This completes and fixes #465

### DIFF
--- a/agents/conversational.go
+++ b/agents/conversational.go
@@ -40,7 +40,7 @@ type ConversationalAgent struct {
 
 var _ Agent = (*ConversationalAgent)(nil)
 
-func NewConversationalAgent(llm llms.LLM, tools []tools.Tool, opts ...CreationOption) *ConversationalAgent {
+func NewConversationalAgent(llm llms.Model, tools []tools.Tool, opts ...CreationOption) *ConversationalAgent {
 	options := conversationalDefaultOptions()
 	for _, opt := range opts {
 		opt(&options)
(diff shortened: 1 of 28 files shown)


tmc (approved):
LGTM

PR #526, opened 2024-01-18
weaviate: honor vectorstores.WithEmbedder

corani:
### embeddings: add adapter for EmbedderClient interface

An adapter for the EmbedderClient interface has been added. This allows
the use of ordinary functions as Embedder Clients. This change provides
flexibility in implementing the EmbedderClient interface, as it allows
any function with the appropriate signature to be used as an
EmbedderClient.

### weaviate: allow embedder override in vectorstores options

The changes allow the embedder to be overridden in the vectorstores options.
This is done by changing the embedder calls in `AddDocuments` and `SimilaritySearch`
functions to use the embedder from the options instead of the store.

The default embedder is still the one from the store, but it can be overwritten
by passing a `vectorstores.WithEmbedder` option.

A new test case 'TestWeaviateWithOptionEmbedder' is added to ensure that the
embedder provided as an option to either `AddDocuments` or `SimilaritySearch` takes
precedence over the one provided when creating the Store.

This change provides more flexibility in choosing the embedder at runtime,
which can be useful in scenarios where the same store is used with
different embedders.

### PR Checklist

### DIFF
--- a/embeddings/embedding.go
+++ b/embeddings/embedding.go
@@ -35,6 +35,15 @@ type EmbedderClient interface {
 	CreateEmbedding(ctx context.Context, texts []string) ([][]float32, error)
 }
 
+// EmbedderClientFunc is an adapter to allow the use of ordinary functions as Embedder Clients. If
+// `f` is a function with the appropriate signature, `EmbedderClientFunc(f)` is an `EmbedderClient`
+// that calls `f`.
+type EmbedderClientFunc func(ctx context.Context, texts []string) ([][]float32, error)
+
+func (e EmbedderClientFunc) CreateEmbedding(ctx context.Context, texts []string) ([][]float32, error) {
+	return e(ctx, texts)
+}
+
 type EmbedderImpl struct {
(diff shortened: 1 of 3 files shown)


embeddings/embedding.go
```diff
@@ -35,6 +35,15 @@ type EmbedderClient interface {
 	CreateEmbedding(ctx context.Context, texts []string) ([][]float32, error)
 }
 
+// EmbedderClientFunc is an adapter to allow the use of ordinary functions as Embedder Clients. If
+// `f` is a function with the appropriate signature, `EmbedderClientFunc(f)` is an `EmbedderClient`
+// that calls `f`.
+type EmbedderClientFunc func(ctx context.Context, texts []string) ([][]float32, error)
```
corani, line 41: Utility that I'm using in the test case, but I figured it is generally useful. Mirrors the standard library's `http.HandlerFunc`

tmc (approved):
LGTM

PR #527, opened 2024-01-18
weaviate: add option to deduplicate documents while adding

corani:
### weaviate: add metadata search function in weaviate store

A new function, MetadataSearch, has been added to the Weaviate store. This
function allows for searching Weaviate based on metadata rather than on
similarity. The function takes a context, the number of documents to return,
and a set of options as parameters. It uses these to create a GraphQL query
which is then sent to the Weaviate server. The response is parsed and returned
as a slice of schema.Document.

In addition to the function itself, a test, TestMetadataSearch, has been added
to verify the functionality of the new MetadataSearch function. The test
creates a new Weaviate store, adds some documents to it, and then uses the
MetadataSearch function to retrieve documents based on their metadata. The
results are then checked to ensure they match the expected output.

The addition of this function allows for more flexible querying of Weaviate,
making it easier to find specific documents based on their metadata.

### weaviate: add deduplication functionality to weaviate

The changes introduce a deduplication function to the vectorstores
options. This function is used when adding documents to the store,
allowing the system to filter out duplicate documents before they
are added. This is particularly useful to prevent wasting time on
creating an embedding when one already exists.

The changes also include a test case for the new deduplication
functionality, ensuring that it works as expected. The test case
adds two documents to the store, one of which is a duplicate. The
test verifies that only the unique document is added to the store.

### PR Checklist

- [X] Name your Pull Request title clearly, concisely, and prefixed with the name of the primarily affected package you changed according to [Good commit messages](https://go.dev/doc/contribute#commit_messages) (such as `memory: add interfaces for X, Y` or `util: add whizzbang helpers`).
- [X] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `Fixes #123`).
- [X] Describes the source of new concepts.
- [X] Contains test coverage for new functions.

### DIFF
--- a/vectorstores/options.go
+++ b/vectorstores/options.go
@@ -1,6 +1,11 @@
 package vectorstores
 
-import "github.com/tmc/langchaingo/embeddings"
+import (
+	"context"
+
+	"github.com/tmc/langchaingo/embeddings"
+	"github.com/tmc/langchaingo/schema"
+)
 
 // Option is a function that configures an Options.
 type Option func(*Options)
@@ -11,6 +16,7 @@ type Options struct {
 	ScoreThreshold float32
 	Filters        any
 	Embedder       embeddings.Embedder
+	Deduplicater   func(context.Context, schema.Document) bool
 }
 
 // WithNameSpace returns an Option for setting the name space.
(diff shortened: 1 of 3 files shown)


tmc (approved):
Lovely, LGTM

PR #549, opened 2024-01-25
llms: Add convenience helper for text parts

tmc:
This adds a basic "TextParts" convenience helper to assemble text-only message inputs.


eliben (commented):
Looks good overall. 

A couple of suggestions:

1. There's already a file to place this in `llms/generatecontent.go` - it's all in the same package so should show on pkg.go.dev just the same; Go generally frowns upon files/packags like "util" or "helpers"
2. You could add a test using this new helper, like I added for GenerateFromSinglePrompt; adding in `googleai_test.go` is easy since the free tier of Google AI doesn't require paying to run tests - just grab a key and run them locally. I can do this later if you prefer to just commit the code first

tmc:
Good points, fixing.

PR #557, opened 2024-01-26
vectorstores: Added an option to PGVector to set the collection metadata

### LINKED ISSUE #556: PGVector: Add support for setting the collection metadata
The PGVector Store struct contains a field called `collectionMetadata` as you [can see here](https://github.com/tmc/langchaingo/blob/main/vectorstores/pgvector/pgvector.go#L46):

```go
// Store is a wrapper around the pgvector client.
type Store struct {
	embedder              embeddings.Embedder
	conn                  *pgx.Conn
	postgresConnectionURL string
	embeddingTableName    string
	collectionTableName   string
	collectionName        string
	collectionUUID        string
	collectionMetadata    map[string]any // No way to set this value
	preDeleteCollection   bool
}
```

However, there is no way to set a value for it.

To fix this, an option should be added to allow you to configure the collection metadata.

When the collection already exists and the metadata gets changed, the database record should be updated.

pieterclaerhout:
### PR Checklist

-----

Added an option to PGVector to set the collection metadata

There was no way to set the collection metadata even though the struct had a field for it.

Therefor, I added an option so you can specify the metadata. If the collection already exists, the metadata gets updated.

Fixes #556

### DIFF
--- a/vectorstores/pgvector/options.go
+++ b/vectorstores/pgvector/options.go
@@ -74,6 +74,12 @@ func WithConn(conn *pgx.Conn) Option {
 	}
 }
 
+func WithCollectionMetadata(metadata map[string]any) Option {
+	return func(p *Store) {
+		p.collectionMetadata = metadata
+	}
+}
+
 func applyClientOptions(opts ...Option) (Store, error) {
 	o := &Store{
 		collectionName:      DefaultCollectionName,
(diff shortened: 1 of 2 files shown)


vectorstores/pgvector/options.go
```diff
@@ -74,6 +74,12 @@ func WithConn(conn *pgx.Conn) Option {
 	}
 }
 
+func WithCollectionMetadata(metadata map[string]any) Option {
```
tmc, line 77: please include comments on any/all exported symbols

tmc (approved):
LGTM but please follow on with a comment

PR #559, opened 2024-01-26
vectorstores: Added a unique constraint for the collection name when using PGVector

### LINKED ISSUE #558: PGVector: Collection name should be unique
When you use PGVector as the vector store and you configure the collection name, the name is not constrained to be unique.

When you look at the code that creates the collection, you will see:

```go
func (s *Store) createOrGetCollection(ctx context.Context) error {
	sql := fmt.Sprintf(`INSERT INTO %s (uuid, name, cmetadata)
		VALUES($1, $2, $3) ON CONFLICT DO NOTHING`, s.collectionTableName)
	if _, err := s.conn.Exec(ctx, sql, uuid.New().String(), s.collectionName, s.collectionMetadata); err != nil {
		return err
	}
	sql = fmt.Sprintf(`SELECT uuid FROM %s WHERE name = $1 ORDER BY name limit 1`, s.collectionTableName)
	return s.conn.QueryRow(ctx, sql, s.collectionName).Scan(&s.collectionUUID)
}
```

Every time this function is called, a new collection with a new UUID will be created using the same name. This makes it impossible to reuse the collection.

To fix this, I suggest we add a constraint on the collection name to make it unique. By doing so, a conflict will happen when you try to create a collection with an existing name. In that case, the existing collection will be reused.

If you compare it with the way the Python version of langchain does this, you'll [see they do this](https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/vectorstores/pgvector.py#L92C1-L112C39):

```py
@classmethod
def get_or_create(
    cls,
    session: Session,
    name: str,
    cmetadata: Optional[dict] = None,
) -> Tuple["CollectionStore", bool]:
    """
    Get or create a collection.
    Returns [Collection, bool] where the bool is True if the collection was created.
    """  # noqa: E501
    created = False
    collection = cls.get_by_name(session, name)
    if collection:
        return collection, created

    collection = cls(name=name, cmetadata=cmetadata)
    session.add(collection)
    session.commit()
    created = True
    return collection, created
```

If a collection with that name exists already, it will be used, otherwise, it will create a new one.

pieterclaerhout:
### PR Checklist

-----

Fixes #558

### DIFF
--- a/vectorstores/pgvector/pgvector.go
+++ b/vectorstores/pgvector/pgvector.go
@@ -132,6 +132,7 @@ func (s Store) createCollectionTableIfNotExists(ctx context.Context) error {
 	name varchar,
 	cmetadata json,
 	"uuid" uuid NOT NULL,
+	UNIQUE (name),
 	PRIMARY KEY (uuid))`, s.collectionTableName)
 	if _, err = tx.Exec(ctx, sql); err != nil {
 		return err
--- a/vectorstores/pgvector/pgvector.go
+++ b/vectorstores/pgvector/pgvector.go
@@ -162,7 +162,7 @@ func (s Store) createEmbeddingTableIfNotExists(ctx context.Context) error {
 	cmetadata json,
 	custom_id varchar,
 	"uuid" uuid NOT NULL,
-	CONSTRAINT langchain_pg_embedding_collection_id_fkey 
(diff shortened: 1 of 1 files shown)


tmc (approved):
LGTM

PR #561, opened 2024-01-26
vectorstores: Added the option to specify a custom id when adding documents using PGVector

### LINKED ISSUE #560: PGVector: add support for setting a custom_id value for a document
In the current implementation of the PGVector store, there is a field called `custom_id` present on the `langchain_pg_embedding`  table. This is currently filled with a UUID on every insert as [you can see here](https://github.com/tmc/langchaingo/blob/43478a1119c8802788fc218fc375cf4f758978d4/vectorstores/pgvector/pgvector.go#L202-L213):

```go
customID := uuid.New().String()
b := &pgx.Batch{}
sql := fmt.Sprintf(`INSERT INTO %s (uuid, document, embedding, cmetadata, custom_id, collection_id)
	VALUES($1, $2, $3, $4, $5, $6)`, s.embeddingTableName)

ids := make([]string, len(docs))
for docIdx, doc := range docs {
	id := uuid.New().String()
	ids[docIdx] = id
	b.Queue(sql, id, doc.PageContent, pgvector.NewVector(vectors[docIdx]), doc.Metadata, customID, s.collectionUUID)
}
return ids, s.conn.SendBatch(ctx, b).Close()
```

This seems rather pointless as it will just give every record an unique custom ID which is what the `uuid` field is already doing.

From my point of view, the value of `custom_id` is something you should be able to specify when you add the document so that you can later on use it for example to detect if a specific embedding already exists or not.

I think that adding an option `WithCustomID` on the `AddDocument` `Options` struct would resolve this.

Adding an index on the `custom_id` field would also be useful to speed up searching on that field.

pieterclaerhout:
### PR Checklist

---- 

Fixes #560


tmc (approved):
LGTM

tmc (approved):
LGTM

tmc:
Sorry I should have merged in a more appropriate order here -- can you fix these conflicts @pieterclaerhout?

schema/documents.go
```diff
@@ -5,4 +5,5 @@ type Document struct {
 	PageContent string
 	Metadata    map[string]any
 	Score       float32
+	CustomID    *string
```
tmc, line 8: I'm thinking we should not use a pointer here and treat the empty string as not provided.

textsplitter/split_documents.go
```diff
@@ -8,33 +8,41 @@ import (
...
-	return CreateDocuments(textSplitter, texts, metadatas)
+	return CreateDocuments(textSplitter, texts, metadatas, customIDs)
 }
 
 // CreateDocuments creates documents from texts and metadatas with a text splitter. If
 // the length of the metadatas is zero, the result documents will contain no metadata.
 // Otherwise, the numbers of texts and metadatas must match.
-func CreateDocuments(textSplitter TextSplitter, texts []string, metadatas []map[string]any) ([]schema.Document, error) {
+func CreateDocuments(
+	textSplitter TextSplitter, texts []string, metadatas []map[string]any, customIDs []*string,
```
tmc, line 34: I don't love adding to the palette list like this. I think we could come up with a cleaner interface here. @eliben curious if you'd like to weigh in here?

eliben: It'd be good to start with a clear explanation of what the PR is trying to accomplish - it seems to be touching much more than just PGVector. Some sort of high-level design would be nice (could be just explained in a PR comment / or in a code comment)

pieterclaerhout: The idea behind the `custom_id` field in the PGVector implementation is that you can use to easily relate back to the source of the embedding. In my case for example, the embeddings I insert are coming from a document list in a different database. Being able to specify the custom ID allows me to accomplish a couple of things:

- I can now easily check if embeddings for a database record in the other database exists already
- It makes it easier to selectively delete embeddings that belong together
- When you do a similarity search, you can use the custom ID to relate it back to the original source

When using PGVector, you could probably do the same by storing that ID in the document metadata, but I'm not entirely sure how well that translates to other vector databases. Having it as a separate column also makes it easier to ensure a database index is present to speed up the searches.

The initial idea was to use a single custom ID for each call to `AddDocuments` on the vector store. However, I realised that this approach was too limited and that it would be better to have a custom ID for each individual document that needs to be added to the vector store.

The reason I added it to the Pinecone implementation as well is because I saw that #448 was trying to fix the exact same problem.

Without the option to specify the custom_id for each embedding, that field is kind of useless in the database. Before this PR, a random UUID would get inserted. This serves no purpose. Each record already has a unique UUID in the `uuid` field anyway. Storing two UUIDs for each embedding with one having no proper purpose is just a waste of database resources, right? If you don't want to use a custom ID, we should just use a null-value in the database.

I'll create an new pull request based on this one with a proposition for a cleaned up interface for the `SplitDocuments` and `CreateDocuments` functions.

schema/documents.go
```diff
@@ -5,4 +5,5 @@ type Document struct {
 	PageContent string
 	Metadata    map[string]any
 	Score       float32
+	CustomID    string
```
eliben, line 8: Starting here, it's not clear why `CustomID` is needed if we already have the `Metadata` field. Can't a custom ID be placed in metadata, if needed?

textsplitter/split_documents.go
```diff
@@ -1,65 +1,36 @@
...
 	"log"
 	"strings"
 
 	"github.com/tmc/langchaingo/schema"
 )
 
-// ErrMismatchMetadatasAndText is returned when the number of texts and metadatas
-// given to CreateDocuments does not match. The function will not error if the
-// length of the metadatas slice is zero.
-var ErrMismatchMetadatasAndText = errors.New("number of texts and metadatas does not match")
```
eliben, line 14: Why are these changes in this PR? Are they essential to support custom ID? Can they be in a separate PR?

eliben (changes requested):

textsplitter/split_documents.go
eliben: Thanks for the clarification - I'm still not sure I understand how all the pieces of this PR relate together, so I'm leaving more specific questions.

PS: The examples should all build now with the latest alpha tag

textsplitter/split_documents.go
pieterclaerhout: These are needed as when you split documents using a text splitter, we need to preserve the custom ID.

That lead me to refactoring the `SplitDocuments` function to make it simpler. After this refactor, this error was no longer possible, hence it was removed.

schema/documents.go
pieterclaerhout: Good point. As there is already a field `custom_id` in the database, I opted for making that field useful.

The other option is that stuff like this goes into the metadata (which is fine as well to me), but then I suggest we remove the `custom_id` column from the database. It doesn't make sense from my point of view to have a column in the database that is filled with random stuff that is unused (it just add clutter). Removing that column is probably even the best way forward as it avoids that similar stuff can be done in different ways.

The only risk we take there is that people might be relying on this column already which would cause existing implementations to break.

If we leave the column in, we should ensure that you can actually use it to do something meaningful.

textsplitter/split_documents.go
eliben: I suggest moving this to a separate PR, which can come either before or after the custom ID PR - up to you. Keeping one PR per topic makes it much easier to do a thorough review

pieterclaerhout:
@eliben @tmc 

The more I think about this, the more I am in favour of removing the whole idea of a `custom_id`. 

Since it can be stored in the metadata, it can be used via the metadata to filter on it, I suggest we close this PR and that I create one that removes the `custom_id` column from the PGVector implementation.

What do you think?

eliben:
> @eliben @tmc
> 
> The more I think about this, the more I am in favour of removing the whole idea of a `custom_id`.
> 
> Since it can be stored in the metadata, it can be used via the metadata to filter on it, I suggest we close this PR and that I create one that removes the `custom_id` column from the PGVector implementation.
> 
> What do you think?

That sounds like a good idea.

Please keep the new PR focused on just one change; if there are ancillary refactorings you stumble upon while working on it, that's great, but these should be split off as separate PRs.

pieterclaerhout:
@eliben I've created a new PR for the removal of the custom ID option (https://github.com/tmc/langchaingo/pull/590)

PR #571, opened 2024-01-27
llms: convenience methods to create Parts

eliben:


eliben:
@tmc PTAL - some additional convenience methods that make it a bit more pleasant to work with creating `Part`s for multi-content messages (no need to repeat field names in struct initialization).

tmc (approved):
Great helpers, thanks @eliben!

PR #572, opened 2024-01-27
examples: add new vertex and googleai examples

eliben:
Remove older vertex-palm examples


eliben:
@tmc PTAL

The CI will fail to build examples because we don't commit `go.work`, but it does run locally.

PR #590, opened 2024-01-30
vectorstores: Removed the custom ID option for embeddings when using PGVector

pieterclaerhout:
### PR Checklist

---

See this discussion for more details:

https://github.com/tmc/langchaingo/pull/561#discussion_r1468797930

### DIFF
--- a/vectorstores/pgvector/pgvector.go
+++ b/vectorstores/pgvector/pgvector.go
@@ -161,15 +161,10 @@ func (s Store) createEmbeddingTableIfNotExists(ctx context.Context, tx pgx.Tx) e
 	embedding vector%s,
 	document varchar,
 	cmetadata json,
-	custom_id varchar,
 	"uuid" uuid NOT NULL,
 	CONSTRAINT langchain_pg_embedding_collection_id_fkey
 	FOREIGN KEY (collection_id) REFERENCES %s (uuid) ON DELETE CASCADE,
 	PRIMARY KEY (uuid))`, s.embeddingTableName, vectorDimensions, s.collectionTableName)
-	if _, err := tx.Exec(ctx, sql); err != nil {
-		return err
-	}
-	sql = fmt.Sprintf(`CREATE INDEX IF NOT EXISTS %s_custom_id ON %s (custom_id)`, s.embeddingTableName, s.embeddingTableName)
 	if _, err := tx.Exec(ctx, sql); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


eliben (commented):
It would be good for other folks who worked on `pgvector` to chime in

@Abirdcfly 

@tmc WDYT?

tmc (approved):
I agree that with the metdata field this can go there, we can revisit if necessary.

Abirdcfly:
> It would be good for other folks who worked on `pgvector` to chime in
> 
> @Abirdcfly
> 
> @tmc WDYT?

The pgvector stores the data in postgresql, IMO, the code is changeable, the data is a bit more precious. 

When creating the table structure, I basically copied the langchain-python structure, and I wanted the table created by langchain-python could be reused by the langchain-go code.

https://github.com/langchain-ai/langchain/pull/1630/files#diff-dc91245f981bb33717e95ed1c7347e5dfbd214a1a243a3f516a0895481a655a5R92 langchain/vectorstores/pgvector.py:92
```python
class EmbeddingStore(BaseModel):
    __tablename__ = "langchain_pg_embedding"

    collection_id: Mapped[UUID] = sqlalchemy.Column(
        UUID(as_uuid=True),
        sqlalchemy.ForeignKey(
            f"{CollectionStore.__tablename__}.uuid",
            ondelete="CASCADE",
        ),
    )
    collection = relationship(CollectionStore, back_populates="embeddings")

    embedding: Vector = sqlalchemy.Column(Vector(ADA_TOKEN_COUNT))
    document = sqlalchemy.Column(sqlalchemy.String, nullable=True)
    cmetadata = sqlalchemy.Column(sqlalchemy.JSON, nullable=True)

    # custom_id : any user defined id
    custom_id = sqlalchemy.Column(sqlalchemy.String, nullable=True)
```

`custom_id` is a field that has existed since the very beginning of the pull request.

langchain-python uses this field by passing in optional parameters:

https://github.com/langchain-ai/langchain/blob/c37ca45825b37eed90d773691772a6be4da4322b/libs/community/langchain_community/vectorstores/pgvector.py#L405

```python
    def add_texts(
        self,
        texts: Iterable[str],
        metadatas: Optional[List[dict]] = None,
        ids: Optional[List[str]] = None,
        **kwargs: Any,
    ) -> List[str]:
        """Run more texts through the embeddings and add to the vectorstore.

        Args:
            texts: Iterable of strings to add to the vectorstore.
            metadatas: Optional list of metadatas associated with the texts.
            kwargs: vectorstore specific parameters

        Returns:
            List of ids from adding the texts into the vectorstore.
        """
        embeddings = self.embedding_function.embed_documents(list(texts))
        return self.add_embeddings(
            texts=texts, embeddings=embeddings, metadatas=metadatas, ids=ids, **kwargs
        )
```

I now think it's possible to expose the filter and use this field using `vectorstores.Option`, and it's fine to create it using another pull request if anyone needs it.

tmc:
Hmm I should have referenced the python implementation. With it as a separate field there I'm more inclined to preserve it.

PR #593, opened 2024-01-31
agents: [OneShotZeroAgent] Change regex to allow multi-line inputs

terwey:
This small code change is to allow the one-shot zero agent to pass multi-line input to a tool. I did not change the conversational one as the regex there is different from this one to start with and is missing unit tests (so I cannot confirm breaking changes).

Change is setting the `s` flag to true on the regex.

### PR Checklist


tmc (approved):
LGTM

Issue #231, opened 2023-08-03
textsplitter.RecursiveCharacter does not work well


dl942702882:
`type LocalDoc2Chunk struct {
}

func (l *LocalDoc2Chunk) Do(ctx context.Context, req *Doc2ChunkRequest) (*Doc2ChunkResponse, error) {
	
	c := &textsplitter.RecursiveCharacter{
		Separators:   []string{"\n\n", "\n", "。", "，", ",", "？", "?", "！", "!", "；", ";", ""},
		ChunkSize:    req.ChunkSize,
		ChunkOverlap: 0,
	}
	ret, err := c.SplitText(req.Content)
	if err != nil {
		return nil, err
	}
	return &Doc2ChunkResponse{Chunks: ret, Identity: req.Identity, ChunkSize: req.ChunkSize}, nil
}`

the result is  very different from langchain python;

FluffyKebab:
Can you provide the result you get  in python and go and the python code you use?

dl942702882:
version: v0.1.3
```
func TestChunks2(t *testing.T) {
	logger := log.DefaultLogger
	_ = utils.LoadConfig(logger)
	c := textsplitter.RecursiveCharacter{
		Separators:   []string{"\n\n", "\n", "，", " "},
		ChunkSize:    24,
		ChunkOverlap: 0,
	}
	a, err := c.SplitText(content2)
	if err != nil {
		t.Error(err)
		return
	}
	fmt.Println(len(a))
	for i := range a {
		fmt.Printf("分段%d: %s\n", i, a[i])
	}
}

var content2 string = `
	天天猫您好，很高，

您好，很高兴为您服务，请问有什么可以帮助您的吗？

`
```

dl942702882:
another problem is that,  you cal the chunk size use  golang len function, but len function  just cal the Byte
but in chinese and use UTF-8 encode， one chinese character use 3-4 Byte； this will lead to chunk size parameter not work very well

tmc:
> another problem is that, you cal the chunk size use golang len function, but len function just cal the Byte but in chinese and use UTF-8 encode， one chinese character use 3-4 Byte； this will lead to chunk size parameter not work very well

Hmm yes, we should do a pass and remove code that is utf8-naive.

PR #609, opened 2024-02-09
feat: add new textspliter option: lenfunc

whyiug:
I want to add a new parameter to the textspliter struct,**lenfunc**, to represent a custom length function. In this way, sentences can be splited according to lenfuc, for example,  ` utf8.RuneCountInString` can be passed to split Chinese characters.
Also aligned with the python code https://github.com/langchain-ai/langchain/blob/00a09e1b7117f3bde14a44748510fcccc95f9de5/libs/langchain/langchain/text_splitter.py.
For me, it's a must-have feature.


whyiug:
Also satisfy issue #231

whyiug:
ping @tmc

tmc:
Hmm, it might be best to just default to RuneCount/RuneCountInString to treat content more naturally, thoughts?

I think I prefer this over erroneously using len() on strings.

tmc:
This is better, but I can't help but think we shouldn't even make this configurable and just use utf8.RuneCountInString -- thoughts?

textsplitter/options.go
```diff
@@ -20,6 +23,7 @@ func DefaultOptions() Options {
 		ChunkSize:    _defaultTokenChunkSize,
 		ChunkOverlap: _defaultTokenChunkOverlap,
 		Separators:   []string{"\n\n", "\n", " ", ""},
+		LenFunc:      defaultLenFunc,
```
tmc, line 26: this could just be `utf8.RuneCountInString` instead of being wrapped with defaultLenFunc, I believe.

whyiug:
Hmm, I don't think so. 
Some scenarios require custom length functions. 
Let's say I want to have the same number of words, not characters  in each chunk.
I can simply set the LenFunc:
```go
func countWords(s string) int {
    words := strings.Fields(s)
    return len(words)
}
```
Or the number of tokens divided by other word dividers.

tmc (approved):
LGTM

PR #612, opened 2024-02-17
tool: sql_database generate sql-query filter redundant text

devinyf:
fix: using sql_database tool generate sql-query sometimes contain unexpected redundant text illusion information
like this:
```md
SQLQuery: SELECT MAX(code) AS max_code FROM example_table;
SQLResult: 3 (illusion data)
Answer: There are 3 data in the table. (illusion data)
```
It should only contain sql syntax as:
```md
SELECT MAX(code) AS max_code FROM example_table;
```

### PR Checklist

- [x]  Read the [Contributing documentation](https://github.com/tmc/langchaingo/blob/main/CONTRIBUTING.md).
- [x]  Read the [Code of conduct documentation](https://github.com/tmc/langchaingo/blob/main/CODE_OF_CONDUCT.md).
- [x]  Name your Pull Request title clearly, concisely, and prefixed with the name of the primarily affected package you changed according to [Good commit messages](https://go.dev/doc/contribute#commit_messages) (such as `memory: add interfaces for X, Y` or `util: add whizzbang helpers`).
- [x]  Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [x]  Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `Fixes #123`).
- [x]  Describes the source of new concepts.
- [x]  References existing implementations as appropriate.
- [x]  Contains test coverage for new functions.
- [x]  Passes all [`golangci-lint`](https://golangci-lint.run/) checks.

PR #614, opened 2024-02-18
Fix agent stream callback

devinyf:
I did a little change in this PR to handled 2 problems when using AgentFinalStreamHandler:

I followed  this example to study using agent
https://github.com/tmc/langchaingo/blob/main/examples/mrkl-agent-example/mrkl_agent.go
and use a handler to get streaming result  like this:
```go
	callbackHandler := callbacks.NewFinalStreamHandler()
	// callbackHandler.PrintOutput = true

	executor, err := agents.Initialize(
		llm,
		agentTools,
		agents.ZeroShotReactDescription,
		agents.WithMaxIterations(5),
		agents.WithCallbacksHandler(callbackHandler),
	)
```

#### problem1. Detect keyword failed:  
reason: cut out slice handler.LastTokens before doing detect keyword will cause Detection can’t success

for example:  
before the handler.LastTokens string value is 
```
"result Final Answer: As of February 18,"
```
after the reassign the slice value, it's become
```
"Answer: As of February 18,"
```
Then it can't pass the keywords check process which should contained `Final Answer:` keyword in LastTokens.

So I changed code order,  check keywords first, then reassign slice

#### problem2. Final Answer sometimes contain  colon symbol `:` at the beginning
Add few code to filter the colon if it exist

May this be a positive contribution.

### PR Checklist

### DIFF
--- a/callbacks/agent_final_stream_test.go
+++ b/callbacks/agent_final_stream_test.go
@@ -1,0 +1,29 @@
+package callbacks
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestFilterFinalString(t *testing.T) {
+	t.Parallel()
+
+	keywork := "Final Answer:"
+	correctStr := "This is a correct final string."
+
+	mockchunkStr1 := fmt.Sprintf(` some other text.
+	Final Answer: %s`, correctStr)
+
+	mockchunkStr2 := fmt.Sprintf(`   :    %s`, correctStr)
+
+	filteredStr := filterFinalString(correctStr, keywork)
+
+	filteredStr1 := filterFinalString(mockchunkStr1, keywork)
+	filteredStr2 := filterFinalString(mockchunkStr2, keywork)
+
+	require.Equal(t, filteredStr, correctStr)
+	require.Equal(t, filteredStr1, correctStr)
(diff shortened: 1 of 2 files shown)


tmc (commented):
can you flesh out some more expected cases and use a table driven test in the test method?

devinyf:
I've updated test case in this PR  and #612 , Both using table driven

callbacks/agent_final_stream_test.go
```diff
@@ -0,0 +1,37 @@
...
+	keywork := "Final Answer:"
+
+	// correct final string
+	correctStr := "This is a correct final string."
+
+	extraStrAbore := fmt.Sprintf(" some other text.\nFinal Answer: %s", correctStr)
+	extraStrBefore := fmt.Sprintf(" another text. Final Answer: %s", correctStr)
+	extraColonWithSpacesBefore := fmt.Sprintf(`   :    %s`, correctStr)
+
+	testCases := []filterFinalStringTestCase{
```
tmc, line 26: please place this at the top of the test (after t.Parallel()) and approach it like:

```go
cases := []struct { ...```

Comment (devinyf) [reply callbacks/agent_final_stream_test.go]: updated all test case inline with:
```go
cases := []struct { ...
```
please review

tmc (approved):
LGTM

Issue #323, opened 2023-10-16
BUG - serpapi client wrong query encoding


adrian-branescu:
### Replication scenario:

* langchaingo version:
```
github.com/tmc/langchaingo v0.0.0-20230929160525-e16b77704b8d
```

* Provide any agent type with a `serpapi` instance as a search engine tool.

* Perform the following query (or any other weather related query):
```
current weather in Ramnicu Valcea, Romania
```

* Result:
```
No good Google Search Results was found
```

### Investigation:
@tmc: I compared `serpapi` Python client implementation with yours and it seems one of the sources of this bug is the following [line](https://github.com/tmc/langchaingo/blob/main/tools/serpapi/internal/client.go#L34) where you replace all spaces with `+` before performing URL encoding.

An additional source for this bug might be the fact that you neither trim quotes, nor escape them in the `action.ToolInput` of an Action returned by an Agent.

You can see both problems in the requests extracted from https://serpapi.com/searches/ :

Below is a truncated snippet of the request made by the `serpapi` Python client:
```json
{
  "search_parameters": {
    "engine": "google",
    "q": "current weather in Ramnicu Valcea, Romania",
    "google_domain": "google.com",
    "hl": "en",
    "gl": "us",
    "device": "desktop"
  },
  "search_information": {
    "query_displayed": "current weather in Ramnicu Valcea, Romania",
    "total_results": 824000
  }
}
```

vs the request made by your `serpapi` Golang client:
```json
{
  "search_parameters": {
    "engine": "google",
    "q": ""current+weather+in+Ramnicu+Valcea,+Romania"",
    "google_domain": "google.com",
    "hl": "en",
    "gl": "us",
    "device": "desktop"
  },
  "search_information": {
    "query_displayed": ""current+weather+in+Ramnicu+Valcea,+Romania"",
    "total_results": 0
  }
}
```

PR #616, opened 2024-02-20
serpapi prevent exact-match google search

devinyf:
I just encountered this error similar with the issue #323:
```
No good Google Search Results was found
```

after the investigation, I found sometimes the llm model(e.g. mistral in ollama) will return `Action Input` value surrounded by double quotes
which will became a exact-match behavior and usually find nothing:
```sh
# this is ok, just get name from search engine and continue the following steps
Action: GoogleSearch
Action Input: Oliva Wilde boyfriend age

# error: No good Google Search Results was found
Action: GoogleSearch
Action Input: "Oliva Wilde boyfriend age"
```

with quotes, the actual query string in serpapi became:
```sh
"q": "\"Oliva Wilde boyfriend age\"",
```

and this also appeared in issue #323 
```
    # python version
    "q": "current weather in Ramnicu Valcea, Romania",
    # langchaingo
    "q": ""current+weather+in+Ramnicu+Valcea,+Romania"",
```

another change is the description of this tool:
 I compared serpapi tool description in langchain python and langchaingo,
 I tried several times, In my use experience, I feel the python version would worked better, but I'm not the expert in prompt engineering , so if this should not be changed, I'm going to delete this modification.
```sh
# langchaingo currently use:
`
"A wrapper around Google Search. "
"Useful for when you need to answer questions about current events. "
"Always one of the first options when you need to find information on internet"
"Input should be a search query."
`

# langchain python serpapi:
"A search engine. Useful for when you need to answer questions about current events. Input should be a search query."
```

---
### PR Checklist

### DIFF
--- a/tools/serpapi/internal/client.go
+++ b/tools/serpapi/internal/client.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"strings"
 )
 
 const _url = "https://serpapi.com/search"
@@ -29,6 +30,12 @@ func New(apiKey string) *Client {
 }
 
 func (s *Client) Search(ctx context.Context, query string) (string, error) {
+	// prevent exact-match search: trim " symbol when it exists on both the left and right sides
+	// exact-match behaviour cause `No Google Search Results was found` error too frequently
+	if strings.HasPrefix(query, "\"") && strings.HasSuffix(query, "\"") {
+		query = strings.TrimPrefix(strings.TrimSuffix(query, "\""), "\"")
+	}
+
 	params := make(url.Values)
 	params.Add("q", query)
 	params.Add("google_domain", "google.com")
--- a/tools/serpapi/serpapi.go
+++ b/tools/serpapi/serpapi.go
@@ -37,11 +37,7 @@ func (t Tool) Name() string {
 }
 
 func (t Tool) Description() string {
-	return `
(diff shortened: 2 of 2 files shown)


tools/serpapi/internal/client.go
```diff
@@ -29,6 +30,12 @@ func New(apiKey string) *Client {
 }
 
 func (s *Client) Search(ctx context.Context, query string) (string, error) {
+	// prevent exact-match search: trim " symbol when it exists on both the left and right sides
+	// exact-match behaviour cause `No Google Search Results was found` error too frequently
+	if strings.HasPrefix(query, "\"") && strings.HasSuffix(query, "\"") {
```
tmc, line 35: would be more clear using ``"`` IMO

tmc (approved):
I like the prompt adjustment but I'm less sure about the quote stripping -- putting this in prevents callers specifying exact-search style queries, right?

tools/serpapi/internal/client.go
devinyf: ok, changed to \`"\`

tmc:
I think it's more appropriate for the caller to simply do a strings.Trim() call head of calling search.

devinyf:
> I think it's more appropriate for the caller to simply do a strings.Trim() call head of calling search.

Yes, that’s more preferable, I merge latest main branch and updated it.

PR #625, opened 2024-02-21
googleai: fix options need add default value

Abirdcfly:
You can find out the problem by first executing the tests added in the current pr.

Let me explain what happened:
When calling `llm` via `chain`, because it goes through the function： https://github.com/tmc/langchaingo/blob/853fc0492d2331518252a357759546abfb1f9fea/chains/options.go#L132-L158 

Before this function, only a certain `ChainCallOption` may have been called, but after this function, it will return the 11 l`lms.CallOptions` that are currently available (although some of them are with empty value in there), so the empty value need to be dealt with at the `llm` level.

I'm **guessing** that openai's handler is in the
https://github.com/tmc/langchaingo/blob/853fc0492d2331518252a357759546abfb1f9fea/llms/openai/internal/openaiclient/completions.go#L49-L70

cc @tmc @eliben

### DIFF
--- a/llms/googleai/googleai.go
+++ b/llms/googleai/googleai.go
@@ -50,6 +50,7 @@ func (g *GoogleAI) GenerateContent(ctx context.Context, messages []llms.MessageC
 	for _, opt := range options {
 		opt(&opts)
 	}
+	g.setCallOptionsDefaults(&opts)
 
 	model := g.client.GenerativeModel(opts.Model)
 	model.SetCandidateCount(int32(opts.CandidateCount))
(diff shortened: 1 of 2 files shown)


tmc (approved):
LGTM

llms/googleai/googleai.go
```diff
@@ -50,6 +50,7 @@ func (g *GoogleAI) GenerateContent(ctx context.Context, messages []llms.MessageC
 	for _, opt := range options {
 		opt(&opts)
 	}
+	g.setCallOptionsDefaults(&opts)
```
eliben, line 53: The defaults were just set a few lines up; this doesn't seem right - we should be setting defaults in one place.

Also I don't understand the logic here -- for example, what if the user requested Temperature = 0 explicitly? That's a valid request, why do we override it with a default now?

eliben:
Sorry, I really wanted to have the opportunity to review this PR before it was merged. I left a comment

llms/googleai/googleai.go
Abirdcfly: > Temperature = 0
I later realized this issue: the current code cannot distinguish whether a zero value means it's unset or intentionally set to zero. Currently, only `topP` and `Temperature` can be set to zero.

The correct way I guess might be to modify `getLLMCallOptions` so that it only changes the passed parameters to `lms.CallOptions`? But in that case, it will affect all LLM... Maybe more testing is needed...

eliben: Please revert this PR until we figure out the right way to address this, and create an issue describing the problem where it can be discussed.

Abirdcfly: > create an issue 

#626 

> revert this PR

IMO, the pr that fixes this problem **completely** can revert this commit before submitting their own fix, the current approach has small bug(Can't set `topP` and `Temperature` to 0 on chain calls), but at least it ensures that google ai is working and can return results, without the current pr, chain call without full ChainCallOption will only return 404...

```
TestLLMChainWithGoogleAI
Entering LLM with messages:
Role: human
Text: What is the capital of France
2024/02/22 10:15:23 googleapi: Error 404:
```

eliben:
I'm going to revert this change in `googleai` for now. Let's keep discussing how to solve the real issue in #626

Issue #626, opened 2024-02-22
googleai options need add default value by chain call


Abirdcfly:
Test code

```go
func TestLLMChainWithGoogleAI(t *testing.T) {
	t.Parallel()
	genaiKey := os.Getenv("GENAI_API_KEY")
	if genaiKey == "" {
		t.Skip("GENAI_API_KEY not set")
	}
	model, err := googleai.New(context.Background(), googleai.WithAPIKey(genaiKey))
	require.NoError(t, err)
	require.NoError(t, err)
	model.CallbacksHandler = callbacks.LogHandler{}

	prompt := prompts.NewPromptTemplate(
		"What is the capital of {{.country}}",
		[]string{"country"},
	)
	require.NoError(t, err)

	chain := NewLLMChain(model, prompt)

	result, err := Predict(context.Background(), chain,
		map[string]any{
			"country": "France",
		},
		WithCallback(callbacks.LogHandler{}),
	)
	require.NoError(t, err)
	require.True(t, strings.Contains(result, "Paris"))
}
```

Let me explain what happened:
When calling `llm` via `chain`, because it goes through the function： https://github.com/tmc/langchaingo/blob/853fc0492d2331518252a357759546abfb1f9fea/chains/options.go#L132-L158 

Before this function, only a certain `ChainCallOption` may have been called, but after this function, it will return the 11 l`lms.CallOptions` that are currently available (although some of them are with empty value in there), so the empty value need to be dealt with at the `llm` level.

I'm **guessing** that openai's handler is in the
https://github.com/tmc/langchaingo/blob/853fc0492d2331518252a357759546abfb1f9fea/llms/openai/internal/openaiclient/completions.go#L49-L70

eliben:
> 
> Let me explain what happened: When calling `llm` via `chain`, because it goes through the function：
> 
> https://github.com/tmc/langchaingo/blob/853fc0492d2331518252a357759546abfb1f9fea/chains/options.go#L132-L158
> 

This doesn't seem right to me, because it means that no option is left on its default value when going through a chain. This is really problematic, because different backends may have different defaults.

I wonder why the chain needs to force-initialize all options like this; maybe that's where the root of the problem is.

eliben:
@FluffyKebab @tmc why do we have `chains/options.go` which duplicates most things in `llms/options.go`?

[...]

PR #628, opened 2024-02-22
chains: add test with GoogleAI

eliben:
Resurrect test from #625, but add explicit settings until #626 is resolved


eliben:
@Abirdcfly PTAL - I added back the test you added in #625 and that I later reverted
Options have to be set manually now until we find a good resolution for #626

PR #631, opened 2024-02-23
llms: fix SSE stream should ignore comment

Abirdcfly:
> A colon as the first character of a line is in essence a comment, and is ignored.

ref: https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#event_stream_format

I haven't seen openai officially use colons in SSE, but I have seen openai-like services use colons to prevent connections from being disconnected, for example: `: ping - 2024-02-23 06:52:12.876102` by LLaMA-Factory


tmc (approved):
we should cover this parsing in test but LGTM

PR #639, opened 2024-02-29
chains: fix add ignore StreamingFunc

Abirdcfly:
Fellow #632 
Update #626

### DIFF
--- a/chains/options.go
+++ b/chains/options.go
@@ -205,6 +205,7 @@ func getLLMCallOptions(options ...ChainCallOption) []llms.CallOption { //nolint:
 	if opts.repetitionPenaltySet {
 		chainCallOption = append(chainCallOption, llms.WithRepetitionPenalty(opts.RepetitionPenalty))
 	}
+	chainCallOption = append(chainCallOption, llms.WithStreamingFunc(opts.StreamingFunc))
 
 	return chainCallOption
 }


Abirdcfly:
kindly ping @eliben, and the main branch look like lost some commit like #631

chains/options.go
```diff
@@ -205,6 +207,9 @@ func getLLMCallOptions(options ...ChainCallOption) []llms.CallOption { //nolint:
 	if opts.repetitionPenaltySet {
 		chainCallOption = append(chainCallOption, llms.WithRepetitionPenalty(opts.RepetitionPenalty))
 	}
+	if opts.streamingFuncSet {
```
eliben, line 210: I'm not sure we need the bool for the `func` options; the default for these is `nil`, which should be ok?

So you could just append the option regardless of the bool?

eliben (commented):
Thanks, left a comment

eliben:
Regarding lost commits, @tmc was fixing this issue last week so he will have the context on it here

chains/options.go
Abirdcfly: Updated.

eliben (approved):

Abirdcfly:
> Regarding lost commits, @tmc was fixing this issue last week so he will have the context on it here

ping @tmc

PR #643, opened 2024-03-03
  embeddings: Add Amazon Bedrock embeddings

sansmoraxz:
Adding [AWS bedrock](https://aws.amazon.com/bedrock/) support with embedding generation for cohere and titan.

### PR Checklist


sansmoraxz:
To test you may use the following:-

```go
package bedrock_test

import (
	"context"
	"fmt"
	"testing"

	. "github.com/tmc/langchaingo/embeddings/bedrock"
)

func TestEmbedQuery(t *testing.T) {
	model, err := NewBedrock(WithModel(MODEL_COHERE_EN))
	if err != nil {
		t.Fatal(err)
	}

	embeddings, err := model.EmbedQuery(context.Background(), "hello world")

	if err != nil {
		t.Fatal(err)
	}

	fmt.Println(len(embeddings))
}

func TestEmbedDocuments(t *testing.T) {
	model, err := NewBedrock(WithModel(MODEL_TITAN_EMBED_G1))
	if err != nil {
		t.Fatal(err)
	}

	embeddings, err := model.EmbedDocuments(context.Background(), []string{"hello world", "goodbye world"})

	if err != nil {
		t.Fatal(err)
	}

	fmt.Println(len(embeddings))
}
```

Have not included in the PR as I saw the API calls were not included in existing test cases.

Might I suggest specialized build tags for API calls?

embeddings/bedrock/bedrock_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+	require.NoError(t, err)
+}
+
+func TestEmbedDocuments(t *testing.T) {
+	t.Parallel()
+	model, err := bedrock.NewBedrock(bedrock.WithModel(bedrock.ModelCohereEn))
+	if err != nil {
+		t.Skip("Skipping test, failed to initialize bedrock model")
+	}
```
sansmoraxz, line 28: This is not working I think, it's picking credentials from somewhere. I need some suggestions. Maybe some new environment variable to be explicitly enabled?

sansmoraxz: Added env `TEST_AWS` if you want to test with AWS default credentials chain.

tmc (approved):
LGTM

embeddings/bedrock/bedrock_test.go
tmc: ehh yeah I'd prefer to check for env var poppulation rather than just skipping on init failure.

PR #653, opened 2024-03-07
change mrkl-prompt to python version

devinyf:
change prompt the same as python version:
Current mrkl-prompt in langchaingo would get some very weird answer when I Don't need to use agent-tools in a agent conversation. 
Seems like It's forced llm to use some tool to get the Final Answer 

And this is langchain python mrkl-prompt:
```python
PREFIX = """Answer the following questions as best you can. You have access to the following tools:"""
FORMAT_INSTRUCTIONS = """Use the following format:
"""
```
This have no problem

---
### PR Checklist

### DIFF
--- a/agents/mrkl_prompt.go
+++ b/agents/mrkl_prompt.go
@@ -9,8 +9,8 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


tmc (approved):
LGTM

PR #658, opened 2024-03-08
add waitgroup in ReadFromEgress function for AgentFinalStreamHandler

devinyf:
The ReadFromEgress function is inside a goroutine already
And in some case, It's need to wait until complete from the caller
So just return a waitGroup pointer that user can choose if need to use `wg.Wait()` or not.

---

### PR Checklist


devinyf:
update:  
found that the old implementation in `for xxx := range handler.egress` will never end...
And will cause goroutine leak
fixed this problem using `HandleAgentFinish` method
```go
go func() {
		defer close(handler.egress)
		for data := range handler.egress {
			callback(ctx, data)
}()
```

callbacks/agent_final_stream.go
```diff
@@ -54,16 +59,41 @@ func (handler *AgentFinalStreamHandler) GetEgress() chan []byte {
...
+) *sync.WaitGroup {
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
 	go func() {
-		defer close(handler.egress)
-		for data := range handler.egress {
-			callback(ctx, data)
+		defer wg.Done()
+
+	FORLOOP:
```
tmc, line 72: it's fairly uncommon to break to a label like this, can you consider alternatives here?

devinyf: Yes, use return is more concisely than break label

tmc (commented):
I like the intent here but why shouldn't the call just block?

PR #670, opened 2024-03-13
This example hangs if the JSON data is either marked as human or AI, it only works for me when marked as system

timmattison:
This example hangs if the JSON data is either marked as human or AI, it only works for me when marked as system. This PR makes it so the example no longer hangs.


tmc (approved):
Thanks, LGTM. Surprised this would result in a hang but SGTM.

corani:
I'm not convinced that `human` vs `system` makes the difference here. With more testing I indeed can get it to hang sometimes (might be on the Ollama side?), but I managed that in either case.

It happens (at least for me) when Ollama isn't responding with the correct format so we retry. After a couple of retries it seems to hang.

Maybe a bit of prompt engineering or backtracking would improve things, but I'm surprised at well that we hang.

PR #672, opened 2024-03-14
feat(milvus): The ID is returned after the data is inserted

chenquan:
reference：#590

### PR Checklist

### DIFF
--- a/vectorstores/milvus/milvus.go
+++ b/vectorstores/milvus/milvus.go
@@ -219,9 +219,19 @@ func (s Store) AddDocuments(ctx context.Context, docs []schema.Document,
 	textCol := entity.NewColumnVarChar(s.textField, texts)
 	metaCol := entity.NewColumnVarChar(s.metaField, metadatas)
 	vectorCol := entity.NewColumnFloatVector(s.vectorField, len(vectors[0]), vectors)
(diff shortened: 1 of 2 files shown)


vectorstores/milvus/milvus.go
```diff
@@ -219,12 +219,23 @@ func (s Store) AddDocuments(ctx context.Context, docs []schema.Document,
...
-	_, err = s.client.Insert(ctx, s.collectionName, s.partitionName, vectorCol, metaCol, textCol)
+	column, err := s.client.Insert(ctx, s.collectionName, s.partitionName, vectorCol, metaCol, textCol)
 	if err != nil {
 		return nil, err
 	}
 
-	return nil, nil
+	col, ok := column.(*entity.ColumnInt64)
+	if !ok {
+		return nil, nil
```
tmc, line 229: if we decide to not consider this an error condition there should be a comment describing the rationale

chenquan: > if we decide to not consider this an error condition there should be a comment describing the rationale
Maybe it should return an error.

tmc:
@chenquan sorry about the delay here, as the milvus vectorstore implementation has shifted, can you rebase this to resolve conflicts?

PR #696, opened 2024-03-20
llms: Add support for using the whisper model to transcribe audio

devalexandre:
### PR Checklist

--------------------------------------------------------------------------------------------------


llms/openai/internal/openaiclient/wisper.go
```diff
@@ -0,0 +1,72 @@
+package openaiclient
```
tmc, line 1: this file should be named "whisper.go"

llms/openai/openaillm.go
```diff
@@ -150,3 +150,12 @@ func (o *LLM) CreateEmbedding(ctx context.Context, inputTexts []string) ([][]flo
 	}
 	return embeddings, nil
 }
+
+// Transcription with audio file.
```
tmc, line 154: this comment is incomplete and should begin with the name of the method

llms/openai/openaillm_option.go
```diff
@@ -40,6 +40,8 @@ type options struct {
 	embeddingModel string
 
 	callbackHandler callbacks.Handler
+	// required when use wisper
```
tmc, line 43: typo

llms/openai/openaillm_option.go
```diff
@@ -128,3 +130,10 @@ func WithResponseFormat(responseFormat ResponseFormat) Option {
 		opts.responseFormat = responseFormat
 	}
 }
+
+// WithLanguage allows setting a custom language.
```
tmc, line 134: I'd expect this comment to hint at the format of these values and/or link to docs.

llms/openai/openaillm.go
```diff
@@ -150,3 +150,16 @@ func (o *LLM) CreateEmbedding(ctx context.Context, inputTexts []string) ([][]flo
 	}
 	return embeddings, nil
 }
+
+// Transcription is a method of the Client type that performs transcription of an audio file.
```
tmc, line 154: This comment needs to start with the method name: TranscribeAudio

llms/openai/multicontent_test.go
```diff
@@ -162,6 +162,20 @@ func TestFunctionCall(t *testing.T) {
...
 
+func TestTranscription(t *testing.T) {
+	t.Parallel()
+	llm := newTestClient(t, WithModel("whisper-1"))
+
+	audioFilePath := "./sample.mp3"
+	_, err := os.Stat(audioFilePath)
+	require.NoError(t, err)
+
+	rsp, err := llm.TranscribeAudio(context.Background(), audioFilePath)
```
eliben, line 173: Does it make sense to think of "transcribe audio" in the context of LLMs? AFAIU Whisper is a distinct model from LLMs like the GPT family.

Is this intended as a one-off method only for openai, or as some general audio transcription interface?

devalexandre: only for openai, but I think it's interesting to include it in the general context, because there are other models that do this, but at the moment it's only for openai

corani: I haven't tried it yet, but this seems interesting as a locally running alternative: https://github.com/JigsawStack/insanely-fast-whisper-api 

It would be cool if we could support something like that, so you can combine it with a Ollama to build some local-only tools.

(So far I've been using https://github.com/Purfview/whisper-standalone-win locally, which is a single-binary wrapper around https://github.com/SYSTRAN/faster-whisper)

devalexandre: We can implement it in other LLMS, the problem is that so far I have only found the standalone version for Windows, but I am looking for other alternatives

tmc:
I agree with @eliben's intuition here, I'm not sure if audio transcription as a concept fits right into our llm namespace. I'm open to exposing this and generalizing over providers but I think it belongs in a different namespace.

documentloaders/whisper.go
```diff
@@ -0,0 +1,228 @@
...
+
+	transcribe, err := c.transcribe(ctx, c.audioFilePath)
+	if err != nil {
+		return nil, err
+	}
+
+	// create a virtual file
+	tmpOutputFile, err := os.CreateTemp("", "*.txt")
+	if err != nil {
+		return nil, fmt.Errorf("erro ao criar arquivo temporário de texto: %w", err)
```
corani, line 104: You might want to translate this to English 😉

documentloaders/whisper.go
```diff
@@ -0,0 +1,228 @@
...
+// It returns the path to the temporary file and any error encountered.
+//
+// nolint
+func downloadFileFromURL(fileURL string) (string, error) {
+	parsedURL, err := url.Parse(fileURL)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse URL: %w", err)
+	}
+
+	// Verificação adicional do esquema pode ser realizada aqui, se necessário
```
corani, line 143: idem

devalexandre:
@tmc some update ?

PR #698, opened 2024-03-20
github: Add docs site previews publishing 

tmc:
This adds docs site preview rendering+publishing for pushes/PRs which should assist in documentation contributors.


tmc:
On further inspection I think we should wait for https://github.com/actions/deploy-pages to (hopefully) support preview actions as the repository is currently set up in the more recent approach vs the "classic" gh-pages based approach (which the action this is trying to use expects/needs).

PR #707, opened 2024-03-21
llms/anthropic: adds full support for messages api

### LINKED ISSUE #695: anthropic: Support anthropic messages api
Anthropic has a more structured completion interface now, we should support it.

https://docs.anthropic.com/claude/reference/migrating-from-text-completions-to-messages

joeychilson:
### PR Checklist

This adds full support for the new messages API including streaming. 
It backwards compatible with client option flag for using message api. 
It also updates the default model to 1.3 as 1.0 isn't supported by either completions or messages API.

Closes #695

### DIFF
--- a/llms/anthropic/anthropicllm.go
+++ b/llms/anthropic/anthropicllm.go
@@ -10,6 +10,7 @@ import (
 	"github.com/tmc/langchaingo/callbacks"
 	"github.com/tmc/langchaingo/llms"
 	"github.com/tmc/langchaingo/llms/anthropic/internal/anthropicclient"
+	"github.com/tmc/langchaingo/schema"
 )
 
 var (
@@ -17,6 +18,12 @@ var (
 	ErrMissingToken  = errors.New("missing the Anthropic API key, set it in the ANTHROPIC_API_KEY environment variable")
 
 	ErrUnexpectedResponseLength = errors.New("unexpected length of response")
+)
+
+const (
+	RoleUser      = "user"
+	RoleAssistant = "assistant"
+	RoleSystem    = "system"
 )
 
 type LLM struct {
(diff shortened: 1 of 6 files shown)


joeychilson:
I updated this to remove the fatals from the stream processing. I think it's better to pass the errors back instead of crashing and just letting the user handle them. This is different than the other LLM implementations like OpenAI which I used as a guideline.

llms/anthropic/internal/anthropicclient/messages.go
```diff
@@ -0,0 +1,297 @@
...
+	if !ok {
+		return response, errors.New("invalid delta field type")
+	}
+	if stopReason, ok := delta["stop_reason"].(string); ok {
+		response.StopReason = stopReason
+	}
+
+	usage, ok := event["usage"].(map[string]interface{})
+	if !ok {
+		return response, errors.New("invalid usage field type")
```
tmc, line 275: while an internal package, generally it's better to define these errors up top so calling code can use errors.Is checks.

tmc (approved):
Great contribution! some nits but they're for the internal client.

llms/anthropic/anthropicllm.go
```diff
@@ -36,9 +43,10 @@ func New(opts ...Option) (*LLM, error) {
 
 func newClient(opts ...Option) (*anthropicclient.Client, error) {
 	options := &options{
-		token:      os.Getenv(tokenEnvVarName),
-		baseURL:    anthropicclient.DefaultBaseURL,
-		httpClient: http.DefaultClient,
+		token:             os.Getenv(tokenEnvVarName),
+		baseURL:           anthropicclient.DefaultBaseURL,
+		httpClient:        http.DefaultClient,
+		useCompletionsAPI: true,
```
tmc, line 49: I'd actually like to default to the messages API

Issue #507, opened 2024-01-10
openai: functions are deprecated in the API, "tools" is the new way


eliben:
Based on the [API docs of the new chat endpoints](https://platform.openai.com/docs/api-reference/chat/create), `function_call` is now deprecated and users are encouraged to use `tool_choice` instead.

The only supported kind of `tool` is still a function, but this leaves space for future expansion.

This may be related to https://github.com/tmc/langchaingo/pull/145 -- if we switch to use https://github.com/sashabaranov/go-openai, we won't have to work as hard to changes in the API

President:
ToolChoice option is not respected!

It should be filled here:
```
req := &openaiclient.ChatRequest{
		Model:            opts.Model,
		StopWords:        opts.StopWords,
		Messages:         chatMsgs,
		StreamingFunc:    opts.StreamingFunc,
		Temperature:      opts.Temperature,
		MaxTokens:        opts.MaxTokens,
		N:                opts.N,
		FrequencyPenalty: opts.FrequencyPenalty,
		PresencePenalty:  opts.PresencePenalty,

		FunctionCallBehavior: openaiclient.FunctionCallBehavior(opts.FunctionCallBehavior),
		Seed:                 opts.Seed,
		Metadata:             opts.Metadata,
	}
```

PR #709, opened 2024-03-21
llms: Implement tool calling, including parallel tool call request support

### LINKED ISSUE #712: llms: Evolve interface to enable multiple tool calls
We currently only support one FuncCall on [llms.ContentChoice](https://pkg.go.dev/github.com/tmc/langchaingo@v0.1.7/llms#ContentChoice) -- this should be extended to enable multiple types of tool calls and multiple calls.

tmc:
This adds tool implementation which generalizes function calling.

This focuses on the openai inference backend. Related filed issues: #716 

Fixes #712, Refs #507

### DIFF
--- a/examples/openai-function-call-example/go.mod
+++ b/examples/openai-function-call-example/go.mod
@@ -4,7 +4,7 @@ go 1.21
 
 toolchain go1.21.4
 
-require github.com/tmc/langchaingo v0.1.7
+require github.com/tmc/langchaingo v0.1.8-0.20240324221257-f3c3fcc40b49
 
 require (
 	github.com/dlclark/regexp2 v1.10.0 // indirect
--- a/llms/openai/openaillm.go
+++ b/llms/openai/openaillm.go
@@ -138,22 +138,18 @@ func (o *LLM) GenerateContent(ctx context.Context, messages []llms.MessageConten
 			}
 		}
 		if c.FinishReason == "tool_calls" {
-			// TODO: we can only handle a single tool call for now.
+			// TODO: we can only handle a single tool call for now, we need to evolve the API to handle multiple tool calls.
 			toolCall := c.Message.ToolCalls[0]
 			choices[i].FuncCall = &schema.FunctionCall{
 				Name:      toolCall.Function.Name,
 				Arguments: toolCall.Function.Arguments,
 			}
 		}
-
 	}
-
 	response := &llms.ContentResponse{Choices: choices}
-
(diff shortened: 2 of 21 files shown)


llms/openai/openaillm.go
```diff
@@ -67,35 +68,74 @@ func (o *LLM) GenerateContent(ctx context.Context, messages []llms.MessageConten
...
+			msg.Role = RoleFunction
+		case schema.ChatMessageTypeTool:
+			msg.Role = RoleTool
+			// Here we extract tool calls from the message and populate the ToolCalls field.
+
+			// parse mc.Parts (which should have one entry of type ToolCallResponse) and populate msg.Content and msg.ToolCallID
+			if len(mc.Parts) != 1 {
+				return nil, fmt.Errorf("expected exactly one part for role %v, got %v", mc.Role, len(mc.Parts))
+			}
+			switch p := mc.Parts[0].(type) {
```
eliben, line 80: This is odd -- seems like we should statically know whether a pointer or not is returned and error out in the other case. Which one is it?

examples/openai-function-call-example/go.mod
```diff
@@ -4,7 +4,7 @@ go 1.21
 
 toolchain go1.21.4
 
-require github.com/tmc/langchaingo v0.1.7
+require github.com/tmc/langchaingo v0.1.8-0.20240324221257-f3c3fcc40b49
```
eliben, line 7: nit: the tag doesn't exist yet so this won't work for others

you're probably testing this locally with a go.work file? so this line can be omitted for now and later all examples can be updated with the script when 0.1.8 is tagged?

tmc: This is pointing at this branch temporarily so this can build ahead of these changes merging.

llms/openai/openaillm.go
tmc: I agree it's a bit odd -- since both of these types satisfy isPart() the part could be either of them, hence checking both. Very much open to suggestions here.

eliben: There are several places in the code where we type-switch on parts, e.g. https://github.com/tmc/langchaingo/blob/main/llms/googleai/googleai.go#L125

and we only handle the non-pointer type

This is part of the documentation - passing plain `TextPart` etc. is expected

If we want to match both pointer and plain everywhere, we should probably add more interfaces

tmc: I updated this switch to only handle non-pointer values.

Issue #720, opened 2024-03-26
add  `keep_separator` in textsplitter.NewRecursiveCharacter


zhangi:
## Proposal
add support of `keep_separator` textsplitter.NewRecursiveCharacter, which is enabled by default in Python implementation. Refer to [RecursiveCharacterTextSplitter](https://api.python.langchain.com/en/latest/character/langchain_text_splitters.character.RecursiveCharacterTextSplitter.html#langchain_text_splitters.character.RecursiveCharacterTextSplitter)

## Issue
without `keep_separator`, the current implementation produces chunks whose length are much shorter than the one specified by options `WithChunkSize` .
```
package main

import (
	"fmt"
	"strings"

	"github.com/pkoukk/tiktoken-go"
	"github.com/tmc/langchaingo/textsplitter"
)

func main() {
	content := strings.Repeat("The quick brown fox jumped over the lazy dog. ", 20)

	tokenEncoder, _ := tiktoken.GetEncoding("cl100k_base")
	options := []textsplitter.Option{
		textsplitter.WithSeparators([]string{" "}),
		textsplitter.WithChunkOverlap(0),
		textsplitter.WithChunkSize(100),
		textsplitter.WithLenFunc(func(s string) int { return len(tokenEncoder.Encode(s, nil, nil)) }),
	}
	splitter := textsplitter.NewRecursiveCharacter(options...)
	texts, _ := splitter.SplitText(content)
	for i, t := range texts {
		fmt.Printf("chunk %d: tokens: %d\n", i, len(tokenEncoder.Encode(t, nil, nil)))
	}
}
```

**output**

> chunk 0: tokens: 50
chunk 1: tokens: 50
chunk 2: tokens: 50
chunk 3: tokens: 50

sawilde:
Since it looks like this has been implemented, should this issue be closed?

PR #721, opened 2024-03-26
feat: add WithKeepSeparator option for RecursiveCharacter

zhangi:
Fix issue #720 
Reference Implementation In Python:  [RecursiveCharacterTextSplitter](https://api.python.langchain.com/en/latest/_modules/langchain_text_splitters/character.html#RecursiveCharacterTextSplitter)

### DIFF
--- a/textsplitter/options.go
+++ b/textsplitter/options.go
@@ -121,7 +121,7 @@ func WithReferenceLinks(referenceLinks bool) Option {
 	}
 }
 
-// WithKeepSeparator set the keep_separator for a text splitter
+// WithKeepSeparator set the keep_separator for a text splitter.
 func WithKeepSeparator(keepSeparator bool) Option {
 	return func(o *Options) {
 		o.KeepSeparator = keepSeparator
--- a/textsplitter/options.go
+++ b/textsplitter/options.go
@@ -7,6 +7,7 @@ type Options struct {
 	ChunkSize         int
 	ChunkOverlap      int
 	Separators        []string
+	KeepSeparator     bool
 	LenFunc           func(string) int
 	ModelName         string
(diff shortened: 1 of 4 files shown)


textsplitter/options.go
```diff
@@ -118,3 +120,10 @@ func WithReferenceLinks(referenceLinks bool) Option {
 		o.ReferenceLinks = referenceLinks
 	}
 }
+
+// WithKeepSeparator set the keep_separator for a text splitter.
```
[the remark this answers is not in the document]
zhangi: updated comments for WithKeepSeparator option.

devalexandre (changes requested):
don't need update go.mod

textsplitter/options.go
```diff
@@ -118,3 +120,14 @@ func WithReferenceLinks(referenceLinks bool) Option {
 		o.ReferenceLinks = referenceLinks
 	}
 }
+
+// WithKeepSeparator sets whether the separators should be kept in the resulting
+// split text or not. When it is set to True, the separators are included in the
+// resulting split text. When it is set to False, the separators are not included
```
devalexandre, line 126: add in description the default value

zhangi:
> don't need update go.mod

can elaborate a bit more on the reason? The go.mod is automatically updated after my running 'go mod tidy'. There seems to be a direct dependency from [here](https://github.com/tmc/langchaingo/blob/main/llms/mistral/mistralmodel.go#L8)

textsplitter/options.go
zhangi: updated the default value in comments.

tmc (approved):
LGTM

Issue #718, opened 2024-03-25
llms: implementing support for nvidia.ai


devalexandre:
I've been testing the nvidia api and I found it very interesting, as it uses the same openai pattern, I'm thinking about implementing a new type for nvidia within the openai backend

tmc:
Please do.

PR #730, opened 2024-03-29
openai: implementing support for NVidia

devalexandre:
### PR Checklist

-----------------------------------------------------------------------------------------------------------------
Reference a Issue #718


tmc (commented):
I just landed #733 and added an example -- can you rebase this?

I'm a little skeptical about adding APITypeNvidia as I feel a different BaseURL should be sufficient in most cases.

devalexandre:
> I'm a little skeptical about adding APITypeNvidia as I feel a different BaseURL should be sufficient in most cases.

I thought about this possibility, but if you just use With BaseUrl , it may work for integration, but the URL for embeddings is different, hence the addition of APIType, otherwise we would have to check if there is the word nvidia in the URL and change the URL for embeddings

devalexandre:
> I just landed #733 and added an example -- can you rebase this?
> 
> I'm a little skeptical about adding APITypeNvidia as I feel a different BaseURL should be sufficient in most cases.

so Nvidia have a embedding URL different of base URL, because it I create a new type

tmc (changes requested):
Yeah still not open to doing this in such a vendor specific manner, I'd be open to a WIth* Option that adds this but I don't think we should hard code more vendors as that will cause a maintenance burden.

devalexandre:
> Yeah still not open to doing this in such a vendor specific manner, I'd be open to a WIth* Option that adds this but I don't think we should hard code more vendors as that will cause a maintenance burden.

okay, how we can use using olny url I'll create a embedding for Nvidia

PR #745, opened 2024-04-01
GH actions: update lint workflow to newer version of Go

eliben:
What motivates this is a newer lint version for local running with 1.22, which fixes the loop range issue and doesn't require spurious lint errors in tests any more.


tmc (approved):
LGTM

eliben:
Thanks for the quick review!

PR #782, opened 2024-04-17
examples: remove 'replace' lines in go.mod files

eliben:
We rely on a global go.work for this - per-example 'replace' lines complicate things


CrazyWr:
Hey, I'm curious for this, `go.work` works fine in local, but the go.work file is ignored when comming, how it works in CI job `build-examples` wIthout `go.work` file, the new module is not yet released, it will not find the new module when `go mod tidy`

eliben:
> Hey, I'm curious for this, `go.work` works fine in local, but the go.work file is ignored when comming, how it works in CI job `build-examples` wIthout `go.work` file, the new module is not yet released, it will not find the new module when `go mod tidy`

Correct. Committing `go.work` files or `replace` lines in a library module isn't a great idea because it can hide serious failures for users (when users import packages, their `go.work` files are not honored).

Therefore, if an an example's code requires yet-unreleased code, this example should wait until a tagged release with the changed functionality appears.

See #783 for an example - it won't work in CI because 0.1.8 doesn't have the API changes yet, so it asks @tmc to make a new release and the PR can be then merged (with a go.mod update)

CrazyWr:
I get it, I should add an example for new module after the new module is released

PR #783, opened 2024-04-17
examples: fix openai toolcall example after refactor

eliben:
After #757 we can remove a bunch of boilerplate conversion code from the example

Updates #753 

@tmc please note that this requires tagging a new version (v0.1.9) since the API changed, so we can point the example to that (rather than using a pseudo-version).


tmc:
Thanks @eliben! I'll get 0.1.9 tagged this afternoon.

tmc:
Tagged v0.1.9, and set up github release as pre-release (will finalize once examples are update).

PR #802, opened 2024-04-22
vectorstores/pgvector: support both pgx and pgxpool.

joeychilson:
### PR Checklist

This PR updates the pgvector vector store to support connections from either pgx or pgxpool. 

This introduces breaking changes. 

- Removed Close func.
- Removed WithConnectionURL from options.
- Require WithConn to be set.

A PGXConn interface was added that contains all the funcs needed by the vector store. These funcs are offered by both pgx.Conn and pgxpool.Pool. The only difference between the two is how they handle closing the connection. This is why Close was removed. Close could be kept but would require checking the connection type to make sure the connection was a pgx.Conn and then close it properly. I think the better move is to remove it since vector store doesn't need this functionality 

This also removed WithConnectionURL because removing the Close func the vector store can't safely close the created connection anymore. I think the better move is just passing a connection to the vector store to use. 

WithConn was updated to accept PGXConn and will error if not set. 

If breaking changes aren't desired I can convert it back to the same and just have the close func check to make sure the conn is pgx.Conn when using the Close func.


tmc (approved):
Lovely! looks great.

PR #846, opened 2024-05-21
readme: add scheme to vectorstores/chroma URL

codefromthecrypt:
Perhaps this worked before, but now if you omit the protocol in the URL, you'll get an error like this:

```
2024/05/21 10:15:43 new: Get "localhost:8000/api/v1/heartbeat": unsupported protocol scheme "localhost"
```

### PR Checklist

- [n/a] Describes the source of new concepts.
- [n/a] References existing implementations as appropriate.
- [n/a] Contains test coverage for new functions.
- [n/a] Passes all [`golangci-lint`](https://golangci-lint.run/) checks.


tmc (approved):
Thanks, LGTM

Issue #171, opened 2023-07-05
Structured parser doesn't parse array


abdallamourad:
Hi, I am trying to parse an array of data using structured parser. The json I am trying to parse is 
```json
{
  "ids": [number]
}
```

I am getting this error `map[] json: cannot unmarshal array into Go value of type string`
I tried to change the data type of the [parser](https://github.com/tmc/langchaingo/blob/main/outputparser/structured.go#L77) to map[string]any and it worked perfectly.

I was wondering if this case is handled somewhere else? Or can we add it to the existing parser?

FluffyKebab:
The current structured parser only works with key value pairs of strings, but maybe the comma separated list output parser works for your case? Having a parser that can handle arbitrary JSON is definitely needed.

abdallamourad:
Thank you @FluffyKebab for your response! 

I was wondering if this feature would benefit the project! I could probably pick it up?

FluffyKebab:
Yes, please do! Here are some references from the other versions: [TS version](https://github.com/hwchase17/langchainjs/blob/main/langchain/src/output_parsers/structured.ts), [Python version](https://github.com/hwchase17/langchain/blob/master/langchain/output_parsers/pydantic.py). What do you think is the best golang version of a ZOD schema in the TS version and pydantic in the python version?

navi504:
A PR was opened that used a better openai client and had the structure output parser, but it wasn't merged ontime and became dead 

The better openai client has a strictest output parser 
https://github.com/sashabaranov/go-openai/blob/1153eb2595d1529927757dd6df4de71faaafde02/jsonschema/json.go#L23C2-L23C2 

I also raised a similar issue if it's the right way to go
https://github.com/tmc/langchaingo/issues/197

navicstein:
A PR was opened that used a better openai client and had the structure output parser, but it wasn't merged ontime and became dead 

The better openai client has a structured output parser 
https://github.com/sashabaranov/go-openai/blob/1153eb2595d1529927757dd6df4de71faaafde02/jsonschema/json.go#L23C2-L23C2 

I also raised a similar issue if it's the right way to go
https://github.com/tmc/langchaingo/issues/197

dallman2:
The openai client that was targeted by the PR was what I was using before deciding to switch to langchain. I would definitely recommend finishing the PR, sasha's openai client is really nice to use.

ewintr:
[...]

Issue #197, opened 2023-07-16
`outputparser.NewStructured` is restrictive & not robust enough 


navicstein:
right now, `outputparser.NewStructured`  is based on an internal struct (https://github.com/tmc/langchaingo/blob/main/outputparser/structured.go#L36) that's restrictive to only strings and output values, but I  think it should be user-defined, a good example is from vector store here (https://github.com/tmc/langchaingo/blob/main/vectorstores/options.go#L30) 

#### Workaround 
for now, I defined a custom prompt with the fields I need with a custom `OutputParser`  and fed the prompt directly to the chain

tmc:
Yeah I agree -- making this more flexible would be welcomed.

PR #856, opened 2024-05-24
outputparser: add `Defined` parser to extract a struct from LLM output

erictse:
This introduces an output parser named 'Defined' as an improvement to the existing output parser 'Structured'.

A developer can define a struct with the proper tags which can generate a description of the desired output format for an LLM. The parser then returns a struct of the given type, populated with the data it extracted. The developer experience parallels what's described in this [Build an Extraction Chain tutorial](https://python.langchain.com/v0.2/docs/tutorials/extraction/).

For best results, use this with the tool calling feature of your LLM.

Fixes https://github.com/tmc/langchaingo/issues/171 and https://github.com/tmc/langchaingo/issues/197

### PR Checklist


erictse:
@tmc I've addressed the linter issues, but the `tagalign` issues remain for my test file. This feature may mean looking into disabling the `tagalign` linter may be more worthwhile. It wants to force reordering the `json:` struct tag key so it appears after `describe:`, reducing readability.

erictse:
@FluffyKebab Hi, do you have any feedback on this PR for me? Haven't heard from anyone yet.

erictse:
@eliben Sorry if I'm barking up the wrong tree. I'm not sure who should be reviewing/accepting PRs. Either way, do you have any feedback on this?

eliben:
@erictse this isn't an area of langchaingo I was working on, so I think you'll have to wait for @tmc to review.

outputparser/defined.go
```diff
@@ -0,0 +1,148 @@
...
+)
+
+// Defined parses JSON output from an LLM into Go structs. A properly tagged
+// struct can generate TypeScript interfaces to help LLMs generate responses
+// that follow the desired JSON structure.
+type Defined[T any] struct {
+	schema string
+}
+
+func NewDefined[T any](source T) (Defined[T], error) {
```
tmc, line 21: please comment exported symbols.

outputparser/defined.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	numTaggedFields := 0
+	for i := 0; i < numFields; i++ {
+		if _, ok := sourceType.Field(i).Tag.Lookup("describe"); ok {
+			numTaggedFields++
+		}
+	}
+	if numTaggedFields == 0 {
+		return empty, errors.New("requires at least 1 field with the struct tag 'describe'")
```
tmc, line 40: I feel as though we should still operate with zero tagged fields.

outputparser/defined.go
```diff
@@ -0,0 +1,148 @@
...
+		result.Write(data)
+	default:
+		return empty, fmt.Errorf("unable to marshal '%s' field type", sourceType.Kind())
+	}
+	return Defined[T]{result.String()}, nil
+}
+
+var _ schema.OutputParser[any] = Defined[any]{}
+
+func (p Defined[T]) GetFormatInstructions() string {
```
tmc, line 59: please comment on exported symbols.

outputparser/defined_test.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func TestDefined(t *testing.T) {
+	t.Parallel()
+	type Shape struct {
+		Name     string `json:"shapeName" describe:"shape name"`
+		NumSides int    `json:"numSides" describe:"number of sides"`
+	}
+
+	tests := map[string]struct {
```
tmc, line 37: please include more test cases (separate PR ok) that include wantErr: true cases

tmc (commented):
Generally LGTM but have a few comments.

outputparser/defined.go
erictse: That seems reasonable. I've updated the code, tests, and doc comments accordingly.

erictse:
@tmc I've addressed your comments, including allowing untagged struct fields.

tmc:
I like this broadly, but can’t help but think this represents a limited subset of what we might encode via https://github.com/invopop/jsonschema — which could be reused to handle tool reprs.

thoughts?

erictse:
Using struct tags to generate JSON Schema looks promising! It should probably be on the roadmap.

However, here are reasons for not dropping this pull request:

* More is better (in the unpredictable world of LLMs). Some models may lack or be weaker at JSON Schema support than others. The developer can decide which structured data extraction approach is best for their situation. 
* Easier to use. JSON Schema is more complex, which can take folks longer to work with. This feature is comparable to Pydantic classes, and for developers coming from Python, the concept would be familiar.
* Feature parity. LangChain supports both Pydantic classes and JSON Schema. See [LangChain doc](https://python.langchain.com/v0.2/docs/how_to/structured_output/).
* Faster/cheaper(?) This is speculative, but JSON Schema is more verbose and complex–if it does happen to be more resource intensive for some models, having another approach is helpful.

tmc:
That's all fair, let's do it!

Can you follow on with an example and some docs?

Issue #353, opened 2023-11-12
Why does the Executor.Call function not pass the opts parameter to Agent.Plan?


tyloafer:
When I follow the test to use agent

~~~
func TestMRKL(t *testing.T) {
	t.Parallel()

	if openaiKey := os.Getenv("OPENAI_API_KEY"); openaiKey == "" {
		t.Skip("OPENAI_API_KEY not set")
	}
	if serpapiKey := os.Getenv("SERPAPI_API_KEY"); serpapiKey == "" {
		t.Skip("SERPAPI_API_KEY not set")
	}

	llm, err := openai.New()
	require.NoError(t, err)

	searchTool, err := serpapi.New()
	require.NoError(t, err)

	calculator := tools.Calculator{}

	a, err := agents.Initialize(
		llm,
		[]tools.Tool{searchTool, calculator},
		agents.ZeroShotReactDescription,
	)
	require.NoError(t, err)

        result, err := chains.Run(context.Background(), a, "If a person lived three times as long as Jacklyn Zeman, how long would they live",
		chains.WithTemperature(1),
		chains.WithTopK(1),
		chains.WithTopP(1)) //nolint:lll
	require.NoError(t, err)

	require.True(t, strings.Contains(result, "210"), "correct answer 210 not in response")
}
~~~

I find that i cannot paas param like StopWords/MaxTokens to openai llm

when i follow the code, i find that the Executor.Call function not pass the opts parameter to Agent.Plan
which causes that the param i have set is useless

~~~
unc (e Executor) Call(ctx context.Context, inputValues map[string]any, _ ...chains.ChainCallOption) (map[string]any, error) { //nolint:lll
	inputs, err := inputsToString(inputValues)
	if err != nil {
		return nil, err
	}
	nameToTool := getNameToTool(e.Tools)

	steps := make([]schema.AgentStep, 0)
	for i := 0; i < e.MaxIterations; i++ {
		var finish map[string]any
		steps, finish, err = e.doIteration(ctx, steps, nameToTool, inputs)
		if finish != nil || err != nil {
			return finish, err
		}
	}

	return e.getReturn(
		&schema.AgentFinish{ReturnValues: make(map[string]any)},
		steps,
	), ErrNotFinished
}

func (e Executor) doIteration(
	ctx context.Context,
	steps []schema.AgentStep,
	nameToTool map[string]tools.Tool,
	inputs map[string]string,
) ([]schema.AgentStep, map[string]any, error) {
[...]

PR #857, opened 2024-05-25
agents(#353): add CallOptions to Plan()

disk0Dancer:
Hi! I saw Ann issue during usage package and made implementation.)

I have fixed problem for almost all agent types but **not for OpenAI**.
I do not really understand how to work with theirs API, have no keys..

What do you think about solution?


agents/openai_functions_agent.go
```diff
@@ -67,10 +68,12 @@ func (o *OpenAIFunctionsAgent) functions() []llms.FunctionDefinition {
 }
 
 // Plan decides what action to take or returns the final result of the input.
+// TODO(fix): add {options ...chains.ChainCallOption} to llm request.
```
tmc, line 71: why not implement this here?

PR #865, opened 2024-05-30
feat(googleai): add support for array type in convertToolSchemaType

johnknott:
🔧 Added support for 'array' type in the convertToolSchemaType function to map it to genai.TypeArray. This enhances the functionality by enabling correct type handling for array types.

- Improved the type conversion logic by including the 'array' case.
- Ensures that 'array' types are not set to TypeUnspecified but correctly set to TypeArray.
- With this change you can now include Arrays in your Google JSON Schema tool definitions without error, like you could already with other LLMs


tmc (approved):
LGTM

PR #879, opened 2024-06-11
fix: handle nil pointer exception in chat_prompt_template

ankitraj2330:
There may be situation where promptValue is nil which will cause panic while accessing promptValue.Messages()

```
func (p ChatPromptTemplate) FormatMessages(values map[string]any) ([]llms.ChatMessage, error) {
	promptValue, err := p.FormatPrompt(values)
	if promptValue == nil {
		return nil, err
	}
	return promptValue.Messages(), err
}
```

### PR Checklist


ankitraj2330:
@tmc Can you please review this small bug fix. Thank you!

tmc (approved):
Generally we should wrap errors but yes this LGTM.

PR #880, opened 2024-06-12
llms: Add improved yaml and json marshaling

tmc:
This adds standard yaml marshaling of `llm.MessageContent` types with special casing for single-text-part messages.


AugustDev (approved):
Tested - works!

Issue #716, opened 2024-03-24
anthropic: Add tool/function calling support


tmc:
See https://docs.anthropic.com/claude/docs/functions-external-tools#does-claude-have-any-built-in-tools-that-it-knows

PR #882, opened 2024-06-14
anthropic: Add tool/function calling support

lwlee2608:
Implement #716

* add function calling support for anthropic llm
* added `anthropic-tool-use-example`


tmc (commented):
Thanks for this! We shoudl have additional testing present but I like what I see here.

Issue #889, opened 2024-06-15
Ollama JSON format bug


doslindos:
Hello team,

There's apparently known [issue](https://github.com/ollama/ollama/issues/4446) where sometimes the LLM sends junk whitespaces when mode is set to JSON with Ollama.

How this is related to langchaingo ?
I was trying the nice [example](https://github.com/tmc/langchaingo/blob/main/examples/ollama-functions-example/ollama_functions_example.go) to add function calls to Ollama models. The example itself works fine, but when I added an own function to be called, I started getting this:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x6569b4]

goroutine 1 [running]:
github.com/tmc/langchaingo/llms/ollama.(*LLM).GenerateContent(0xc0000f4280, {0x7505f0, 0x9695e0}, {0xc0000a05a0, 0x2, 0xc0000160f0?}, {0x0, 0x0, 0xc0000ae060?})
	/home/juuso/go/pkg/mod/github.com/tmc/langchaingo@v0.1.10/llms/ollama/ollamallm.go:152 +0x9d4
main.main()
```

I  added some logs and saw that there was indeed correct JSON format and actually correct function call with valid input out from the LLM, but a lot of junk whitespaces.
But there were so many whitespaces that the response `done` field was false.
```
2024/06/15 17:06:46 {codestral:22b-v0.1-q8_0 2024-06-15 15:06:46.237829539 +0000 UTC 0xc000094540 false {0s 0s 0 0s 0 0s}}
```

After I found that [here](https://github.com/tmc/langchaingo/blob/main/llms/ollama/ollamallm.go#L132)  there's an assumption that if stream is false then the response will be "done" everytime and the `resp.Message` is not populated which is later tried to reference in [here](https://github.com/tmc/langchaingo/blob/main/llms/ollama/ollamallm.go#L152). 

I fixed this by changing the clause to: 
```
if *req.Stream == false || response.Done {
```

Which will catch the unfortunate scenario.

There are some open PRs to fix this on Ollama side, but adding this small change wouldn't make a lot of harm IMO and would make it so that expanding from the nicely written example will not turn into an hour debug session and sanity checking for some JSON string format incompatabilities :D

tmc:
Very supportive -- do you want to open a PR for this?

doslindos:
Sure I can slap it in tomorrow 👍

[...]

PR #892, opened 2024-06-16
ollama: Fix JSON format bug issue when not streaming

doslindos:
Fix issue #889 

### PR Checklist

### DIFF
--- a/llms/ollama/internal/ollamaclient/types.go
+++ b/llms/ollama/internal/ollamaclient/types.go
@@ -49,7 +49,7 @@ type Message struct {
 type ChatRequest struct {
 	Model     string     `json:"model"`
 	Messages  []*Message `json:"messages"`
-	Stream    *bool      `json:"stream,omitempty"`
+	Stream    bool       `json:"stream,omitempty"`
 	Format    string     `json:"format"`
 	KeepAlive string     `json:"keep_alive,omitempty"`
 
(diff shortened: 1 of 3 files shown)


tmc:
Thanks! I simplified things a bit and made the example a bit more robust.

llms/ollama/internal/ollamaclient/types.go
```diff
@@ -49,7 +49,7 @@ type Message struct {
 type ChatRequest struct {
 	Model     string     `json:"model"`
 	Messages  []*Message `json:"messages"`
-	Stream    *bool      `json:"stream,omitempty"`
+	Stream    bool       `json:"stream,omitempty"`
```
corani, line 52: This makes the `omitempty` meaningless, because bools can't be "empty".

tmc: that's not quite right, in Go json encoding: `The "omitempty" option specifies that the field should be omitted from the encoding if the field has an empty value, defined as false, 0, a nil pointer, a nil interface value, and any empty array, slice, map, or string.`

corani: hmm, then perhaps I'm mistaken... I recall seeing `"field": false` in JSON output despite `omitempty`, but I haven't tested it so I may be misremembering :)

Issue #890, opened 2024-06-15
Unable to Use LLaMA Model with Groq API in langchaingo


Ashad001:
### Description
I am trying to interact with the LLaMA model `llama3-8b-8192` using Groq. However, I'm encountering issues with the integration, and it seems like the library might be prompting for an OpenAI API key instead of using the Groq API key.

### Steps to Reproduce
1. Initialize the Groq client with the LLaMA model and Groq API Url as specified in this [example](https://github.com/tmc/langchaingo/blob/587d3ce5c4f3479947f9b1dc2d17a7d41b20a6f5/examples/groq-completion-example/groq_completion_example.go#L14). 
2. Attempt to generate a response from the model.

### Expected Behavior
The LLaMA model should generate a response based on the prompt provided using the Groq API.

### Actual Behavior
The library seems to be prompting for an OpenAI API key instead of using the Groq API key.
![image](https://github.com/tmc/langchaingo/assets/93534298/0c6bf201-a016-4936-bdfa-7561a6ad2c02)

### Environment
langchaingo version: v0.1.10
Go version: go version go1.22.3 windows/amd64

### Additional Context
I have confirmed that the API key is correct and the Groq API is accessible. The issue seems to stem from the way langchaingo handles the API keys or model initialization.

Any guidance or support on resolving this issue would be greatly appreciated. Thank you!!!

tmc:
Can you supply a code sample triggering this?

Ashad001:
Apologies for the confusion. I realized I had to add the `openai.WithToken` parameter in `openai.New` method (this [example](https://github.com/tmc/langchaingo/blob/587d3ce5c4f3479947f9b1dc2d17a7d41b20a6f5/examples/groq-completion-example/groq_completion_example.go)) even though I m not using OpenAI but Groq. This was a bit confusing. Perhaps this could be made clearer in the documentation?
Thanks!

tmc:
Definitely need to make this more clear, there should be a dedicated docs page for this I imagine.

Open to making this type of contribution?

Ashad001:
I m new to Go, but I'm sure it'll be a worthwhile experience! Will start working on it soon.

PR #893, opened 2024-06-16
examples: Correct API key usage for Groq integration

Ashad001:
Related To: #890

- Added `godotenv` to load environment variables from `.env` file.
- Corrected the initialization of the OpenAI client to include `openai.WithToken(apiKey)` for proper Groq API key usage.
- Updated the example to reflect the correct usage and included necessary error handling for missing API keys.

This change addresses the confusion regarding the API key parameter and ensures proper integration with the Groq API.

I still think there should be a separate module to load models for Groq instead of OpenAI, for better readability.


tmc (approved):
While we don't typically include a dotenv package, LGTM.

PR #899, opened 2024-06-17
bugfix: Anthropic Function Calling 

lwlee2608:
Fixed a bug from previous PR #882 where Anthropic LLM couldn't identify the input arguments of a function call from the message history.


JckHoe:
https://github.com/tmc/langchaingo/pull/908

can review this fix ? that pr resolves a issue with the request

PR #902, opened 2024-06-17
llms/watsonx: update module

h0rv:
The watsonx Go package just got transfered to the official IBM org here: https://github.com/ibm/watsonx-go.

There is also a few API changes along with this.


h0rv:
Build example will fail due to it not pointing to new package yet

tmc:
Great to see!

PR #872, opened 2024-06-08
[llms] Fix ErrIncompleteEmbeddings Message Typos

wgeorgecook:
### PR Checklist

- [x] Read the [Contributing documentation](https://github.com/tmc/langchaingo/blob/main/CONTRIBUTING.md).
- [x] Read the [Code of conduct documentation](https://github.com/tmc/langchaingo/blob/main/CODE_OF_CONDUCT.md).
- [x] Name your Pull Request title clearly, concisely, and prefixed with the name of the primarily affected package you changed according to [Good commit messages](https://go.dev/doc/contribute#commit_messages) (such as `memory: add interfaces for X, Y` or `util: add whizzbang helpers`).
- [x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `Fixes #123`).
- [n/a] Describes the source of new concepts.
- [n/a] References existing implementations as appropriate.
- [n/a] Contains test coverage for new functions.
- [n/a] Passes all [`golangci-lint`](https://golangci-lint.run/) checks.

This PR fixes two typos in the `ErrIncompleteEmbeddings` message. The first is the word `no` instead of `not` being incorrectly used. The second removes an extraneous `m` in the word `embedded`.  I could not find any other PRs for this. I haven't changed any functionality, only the content of the declared error.

PR #925, opened 2024-06-22
llms: extract common llms errors to shared declaration

wgeorgecook:
### PR Checklist

This addresses the comment Travis made [in this PR I previously made](https://github.com/tmc/langchaingo/pull/872#pullrequestreview-2119834685). It specifically extracts the ErrorIncompleteEmbedding error and the ErrorEmptyResponse error used in the files in the `/llms` directory. 

I specifically targeted these errors because they're the ones I encountered using LangChain Go, there is still more work to be done in consolidating all of the other potentially shared errors.

### DIFF
--- a/llms/errors.go
+++ b/llms/errors.go
@@ -3,9 +3,9 @@ package llms
 import "errors"
 
 var (
-	// ErrEmptyResponse is thrown when an LLM returns an empty response
+	// ErrEmptyResponse is thrown when an LLM returns an empty response.
 	ErrEmptyResponse = errors.New("no response")
 	// ErrIncompleteEmbedding is thrown when the length of an embedding
-	// request does not match the length of the returned embeddings array
+	// request does not match the length of the returned embeddings array.
 	ErrIncompleteEmbedding = errors.New("not all input got embedded")
 )
--- a/llms/errors.go
+++ b/llms/errors.go
@@ -3,5 +3,9 @@ package llms
 import "errors"
 
 var (
+	// ErrEmptyResponse is thrown when an LLM returns an empty response
+	ErrEmptyResponse = errors.New("no response")
+	// ErrIncompleteEmbedding is thrown when the length of an embedding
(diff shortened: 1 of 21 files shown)


wgeorgecook:
Looks like the ci lint wasn't run before these files made it into main. I'm working on fixing them all now.

wgeorgecook:
CI is now passing.

wgeorgecook:
@tmc I resolved conflicts with main for this and ensured that the linter is happy in the `/llms` directory. Let me know if you want to take this in a different direction but I think this is a good start towards centralizing these error messages.

tmc:
Thanks so much, this ia great to see.

llms/errors.go
```diff
@@ -0,0 +1,11 @@
+package llms
+
+import "errors"
+
+var (
+	// ErrEmptyResponse is thrown when an LLM returns an empty response.
```
tmc, line 6: "thrown" isn't exactly the right language in Go,

tmc:
I wonder if llms is the right place to put the errors, I don't love the new source dependency on what is already a pretty large package implies. I don't have a better suggestion. @eliben do you have an opinion here?

llms/errors.go
wgeorgecook: @tmc Fair. I'll change it to "is returned"

wgeorgecook:
@tmc I have the same concern. It feels weird to have a root level errors package since the exported types _are_ directly related to the LLMs. But it doesn't have to be. But it would probably also feel weird to have a root level errors package with just any and all exported errors in it. Very open to suggestions.

eliben:
> I wonder if llms is the right place to put the errors, I don't love the new source dependency on what is already a pretty large package implies. I don't have a better suggestion. @eliben do you have an opinion here?

I don't see the advantage of this approach - what's the motivation here? How are llm-unique errors going to be represented?

PR #935, opened 2024-06-27
llms: Add fake package

devalexandre:
### PR Checklist


devalexandre:
> You checked " Provide a description in this PR that addresses what the PR is solving, or reference the issue that it solves (e.g. Fixes #123).", but in fact it's not clear at all what and why this is doing.
> 
> Also, the code is full of Portuguese comments and strings, is this intentional?

sorry , I'll update to en and edit the list

flc1125:
I suggest adding a `Generate` interface to pass in, instead of `responses []string`.

Then implement a `Default Generate` to support it. This would allow for better customization and extensibility.

tmc (approved):
LGTM. would be cool to support streaming.

Please leave source comments in english

tmc (approved):
LGTM

PR #950, opened 2024-07-07
anthropic: add image read and stream tool use ability

diemus:
the current version does not implements image reading and stream tool use , so I add it


tmc (changes requested):
Thanks for this contribution, this would need tests and the PR has lint issues.

Issue #948, opened 2024-07-05
Ollama Package: Not handling context timeout issues on streaming response.


ardan-bkennedy:
When a context timeout occurs at the time a response is being provided, this line of code throws a panics. It appears `resp.Message` is  `nil` and it's not being checked.

https://github.com/tmc/langchaingo/blob/main/llms/ollama/ollamallm.go#L152

```
goroutine 1 [running]:
github.com/tmc/langchaingo/llms/ollama.(*LLM).GenerateContent(0x14000444140, {0x104a823f0, 0x14000092000}, {0x140003044e0, 0x1, 0x104dc0f18?}, {0x14000320050, 0x1, 0x14000320050?})
	/Users/bill/code/go/src/github.com/ardanlabs/ai-training/vendor/github.com/tmc/langchaingo/llms/ollama/ollamallm.go:159 +0x7ec
github.com/tmc/langchaingo/llms.GenerateFromSinglePrompt({0x104a823f0, 0x14000092000}, {0x104a80b80, 0x14000444140}, {0x14000493980, 0x1974}, {0x14000320050, 0x1, 0x1})
	/Users/bill/code/go/src/github.com/ardanlabs/ai-training/vendor/github.com/tmc/langchaingo/llms/llms.go:41 +0x148
github.com/tmc/langchaingo/llms/ollama.(*LLM).Call(...)
	/Users/bill/code/go/src/github.com/ardanlabs/ai-training/vendor/github.com/tmc/langchaingo/llms/ollama/ollamallm.go:44
main.questionResponse({0x104a823f0, 0x14000092000}, {0x14000136168, 0x17}, {0x14000432000, 0x5, 0x14000114600?})
```

nickisworking:
i had the similar issue,  #961 may solve this

ardan-bkennedy:
It's not that I couldn't solve the problem with a larger timeout. This is something that needs to be fixed in the code so when a timeout occurs it happens gracefully.

ardan-bkennedy:
Nice, you provided a PR to fix it. very cool.  Thanks.  At this point I have a properly set timeout.

PR #961, opened 2024-07-23
ollamaclient: add scanner error checking in stream function for handling context cancellation

nickisworking:
When using `ollamaclient`, if the application triggers a context cancel for some reason(i.e. handling user cancellation in case of chatbot), and the cancel signal propagates before making the HTTP call (146 line in ollamaclient.go), an error is returned correctly.

However, if the cancel signal propagates after the HTTP call and response.Body is closed, a nil pointer exception occurs at line 152 of `ollamallm.go` because the scanner loop is not entered and ollamaclient.ChatResponse is not created.

Therefore, while I considered pre-creating ollamaclient.ChatResponse in the oolamallm package, I decided that propagating the context.Canceled error would be a better approach.

### PR Checklist

### DIFF
--- a/llms/ollama/internal/ollamaclient/ollamaclient.go
+++ b/llms/ollama/internal/ollamaclient/ollamaclient.go
@@ -180,6 +180,10 @@ func (c *Client) stream(ctx context.Context, method, path string, data any, fn f
 		}
 	}
 
+	if err := scanner.Err(); err != nil {
+		return err
+	}
+
 	return nil
 }
 


nickisworking:
this would fix this issue #948

nickisworking:
@tmc can I get a review? 👀

aria3ppp:
@tmc it is a simple fix in a situation when context deadline triggers and then cause a segmentation fault on `resp.Message.Content`.

Issue #956, opened 2024-07-12
docker-compose fails in pgvector-vectorstore-example with "The Compose file is invalid" error


mlliarm:
Hello,

I'm trying to run the example and I'm getting the following:

```
$ docker-compose up
ERROR: The Compose file is invalid because:
Service db has neither an image nor a build context specified. At least one must be provided.
```

After [a bit of googling](https://stackoverflow.com/questions/74606644/running-docker-compose-up-ends-up-with-error-service-has-neither-an-image-nor-a) I found out that the context is missing from the build in the `docker-compose.yml` file.

This file currently is:

```yml
version: "3.9"
services:
  db:
    build:
      dockerfile: postgres.Dockerfile
    restart: always
    ports:
      - "5432:5432"
    environment:
      POSTGRES_PASSWORD: testpass
      POSTGRES_USER: testuser
      POSTGRES_DB: testdb
```

When changed to the following `docker-compose.yml` according to [this suggestion](https://stackoverflow.com/a/74606959)

```yml
version: "3.9"
services:
  db:
    build:
      context: .
      dockerfile: postgres.Dockerfile
    restart: always
    ports:
      - "5432:5432"
    environment:
      POSTGRES_PASSWORD: testpass
      POSTGRES_USER: testuser
      POSTGRES_DB: testdb
```
it works:

```bash
$ docker-compose up
Creating network "pgvector-vectorstore-example_default" with the default driver
Building db
DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            Install the buildx component to build images with BuildKit:
            https://docs.docker.com/go/buildx/

Sending build context to Docker daemon   34.3kB
Step 1/4 : FROM postgres:16
16: Pulling from library/postgres
f11c1adaa26e: Pull complete 
76ce212b9153: Pull complete 
919ca406a058: Pull complete 
...
```

If you agree I can send a PR with this change.

Cheers.

elnoro:
Hello,

According to the documentation here: https://docs.docker.com/compose/compose-file/build/#context, the context should already default to ".":

> If not set explicitly, context defaults to project directory (.).

However, Compose V1 ("docker-compose") has reached its end-of-life and is no longer supported by Docker. It might be a good idea to update the README to use Compose V2 ("docker compose" without the dash).

Thank you!

PR #969, opened 2024-07-27
examples: replaced compose v1 with compose v2 in readme for pgvector example

elnoro:
A minor issue with pgvector example README:

https://docs.docker.com/compose/migrate/

Compose V1 used to be available as `docker-compose`. It reached its end of life last year.
Compose V2 is available as `docker compose` (no dash), and if you don't have an alias, the example will fail.

Related to https://github.com/tmc/langchaingo/issues/956


tmc (approved):
LGTM

PR #992, opened 2024-08-16
milvus: add skipFlushOnWrite option

fatedier:
When importing a large number of documents at once, flushing during each `AddDocuments` can significantly increase the time taken.

In my scenario, I create a new collection, insert datas, and then replace the original collection, so there is no need to flush every time data is inserted.

Add an option `skipFlushOnWrite` to optimize this scenario. In my test scenario, the total time was reduced from 38 minutes to 5 minutes.

### PR Checklist


tmc (approved):
LGTM - consider linking to docs for this option.

PR #1003, opened 2024-09-05
redisvector: fix score threshold option

acrmp:
* Redis vector range queries [expect a radius](https://redis.io/docs/latest/develop/interact/search-and-query/advanced-concepts/vectors/#vector-range-queries) rather than a score threshold.
* The existing code passed the score threshold through directly as the radius param. This means that a higher score threshold will result in in more matches than a lower score threshold which seems wrong. Instead this PR subtracts the score threshold from 1.
* The langchain project appears to have deprecated score threshold in favour of distance threshold. A future improvement would be to add a `WithDistanceThreshold` option for consistency.
* This PR is a breaking change for clients using redisvector and `WithScoreThreshold`. An alternative would be to keep the existing behaviour but add `WithDistanceThreshold` and guide users towards it.

### PR Checklist

- [X] Name your Pull Request title clearly, concisely, and prefixed with the name of the primarily affected package you changed according to [Good commit messages](https://go.dev/doc/contribute#commit_messages) (such as `memory: add interfaces for X, Y` or `util: add whizzbang helpers`).
- [X] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `Fixes #123`).
- [X] ~~Describes the source of new concepts.~~
- [X] ~~Contains test coverage for new functions.~~
- [X] Passes all [`golangci-lint`](https://golangci-lint.run/) checks. - There are some existing complaints about magic numbers not introduced by this PR.

### DIFF
--- a/vectorstores/redisvector/index_search.go
+++ b/vectorstores/redisvector/index_search.go
@@ -98,19 +98,19 @@ func (s IndexVectorSearch) AsCommand() []string {
 
 	const vectorField = "vector"
 	const vectorFieldAs = defaultDistanceFieldKey
-	const disThresholdFiled = "distance_threshold"
+	const disThresholdField = "distance_threshold"
 	const vectorKey = defaultContentVectorFieldKey
 	params := []string{vectorField, VectorString32(s.vector)}
 
 	if s.scoreThreshold > 0 && s.scoreThreshold < 1 {
 		// Range search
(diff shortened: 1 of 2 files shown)


tmc:
Bit on the fence about a breaking change but this seems more appropriate, in general.

tmc:
Could you tackle the WithDistanceThreshold, that sounds like a nice option.

tmc (approved):
LGTM

Issue #700, opened 2024-03-21
Integration with mongodb for vectorstore


devalexandre:
Mongodb now has support for vector, I'm thinking about starting an implementation based on what's in the documentation, I believe it could be of great value

https://www.mongodb.com/developer/products/atlas/semantic-search-mongodb-atlas-vector-search/

PR #1005, opened 2024-09-06
vectorstores: add mongovector 

prestonvasquez:
resolves https://github.com/tmc/langchaingo/issues/700
[GODRIVER-3305](https://jira.mongodb.org/browse/GODRIVER-3305)

The goal of this PR is to provide a way for users to read and write to an Atlas cluster as a [vector database](https://en.wikipedia.org/wiki/Vector_database) using the MongoDB Go Driver.

A Store should be a wrapper for mongo.Collection, since adding and searching vectors is collection-specific. In this case AddDocuments() and SimilaritySearch() become analogues for Collection.InsertMany() and Collection.Aggregate(). This also keeps our implementation inline with the [Python driver implementation](https://github.com/langchain-ai/langchain/blob/f4ffd692a3548681b4d05c0f255a9b83c07a2e0b/libs/partners/mongodb/langchain_mongodb/vectorstores.py#L206).

Because a single collection can have multiple vector search indexes, we must provide an option to use a specific embedding when adding a document. However, to ensure there is at least a default behavior [where a user does not have to include an optional embedder] we should require that a default embedder be included at construction. 

We've chosen to use `vectorstores.Options.NameSpace` to allow users to change the index on a per-operation basis.

This PR suggests mocking the embedding model to the specifications outlined in this blog post: https://dev.to/prestonvasquez/mocking-an-llm-embedder-targeting-mongodb-atlas-1glp

Integration testing will require setting up a [free tier Atlas Cluster](https://www.mongodb.com/docs/atlas/tutorial/deploy-free-tier-cluster/). The first time the tests run, the vector search indexes will be created. This takes a few minutes. Subsequent tests to the same cluster will not have this requirement.

### PR Checklist

### DIFF
--- a/vectorstores/mongovector/mock_embedder.go
+++ b/vectorstores/mongovector/mock_embedder.go
@@ -104,6 +104,7 @@ func flushEmbedder(ctx context.Context, store Store, emb *mockEmbedder) error {
 }
 
 // newNormalizedFloat32 will generate a random float32 in [-1, 1].
+// nolint:mnd
 func newNormalizedFloat32() (float32, error) {
 	max := big.NewInt(1 << 24)
 
@@ -116,12 +117,14 @@ func newNormalizedFloat32() (float32, error) {
 }
 
 // dotProduct will return the dot product between two slices of f32.
-func dotProduct(v1, v2 []float32) (sum float32) {
+func dotProduct(v1, v2 []float32) float32 {
+	var sum float32
+
 	for i := range v1 {
 		sum += v1[i] * v2[i]
 	}
 
-	return
+	return sum
 }
 
 // linearlyIndependent true if the vectors are linearly independent.
@@ -185,7 +188,7 @@ func dotProductNormFn(score float32, qvector, basis []float32) []float32 {
 	var sum float32
 
 	// Populate v2 upto dim-1.
-	for i := 0; i < len(qvector)-1; i++ {
+	for i := range qvector[:len(qvector)-1] {
 		sum += qvector[i] * basis[i]
 	}
 
--- a/vectorstores/mongovector/mock_embedder.go
+++ b/vectorstores/mongovector/mock_embedder.go
@@ -86,7 +86,8 @@ func (emb *mockEmbedder) EmbedQuery(context.Context, string) ([]float32, error) 
 	return emb.queryVector, nil
 }
 
-func flushEmbedder(ctx context.Context, store Store, emb *mockEmbedder) error {
+// Insert all of the mock documents collected by the embedder.
+func flushMockDocuments(ctx context.Context, store Store, emb *mockEmbedder) error {
 	docs := make([]schema.Document, 0, len(emb.docs))
 	for _, doc := range emb.docs {
 		docs = append(docs, doc)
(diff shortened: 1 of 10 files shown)


vectorstores/mongovector/mongovector.go
```diff
@@ -0,0 +1,221 @@
...
+	coll     mongo.Collection
+	embedder embeddings.Embedder
+	index    string
+	path     string
+}
+
+var _ vectorstores.VectorStore = &Store{}
+
+// New returns a Store that can read and write to the vector store.
+func New(coll mongo.Collection, embedder embeddings.Embedder, opts ...Option) Store {
```
prestonvasquez, line 41: This should accept a pointer to a collection for seamless usability 

```go
	coll := client.Database("vdb").Collection("vstore")
	store := mongovector.New(*coll, nil, nil)
```

prestonvasquez:
@jtazin @matthewdale @qingyang-hu @blink1073

tmc:
Sorry about the delay here, I'll very likely be able to review this today and have been meaning to cut a new release so this should make it in soon!

tmc:
Fantastic change! I'm curious what the plan around the go.mongodb.org/mongo-driver/v2 beta tag is -- I don't love depending on a not-declared-stable package version.

vectorstores/mongovector/mock_embedder.go
```diff
@@ -0,0 +1,207 @@
+package mongovector
```
tmc, line 1: can you move these mocks into a separate subpackage (mocks? mongovectormocks?)

vectorstores/mongovector/mongovector.go
```diff
@@ -0,0 +1,250 @@
+package mongovector
```
tmc, line 1: please include a doc.go file in new packages with a package-level comment

vectorstores/mongovector/mongovector_test.go
```diff
@@ -0,0 +1,592 @@
...
+	"github.com/stretchr/testify/require"
+	"github.com/tmc/langchaingo/embeddings"
+	"github.com/tmc/langchaingo/schema"
+	"github.com/tmc/langchaingo/vectorstores"
+	"go.mongodb.org/mongo-driver/v2/bson"
+	"go.mongodb.org/mongo-driver/v2/mongo"
+	"go.mongodb.org/mongo-driver/v2/mongo/options"
+)
+
+// Run the test without setting up the test space.
```
tmc, line 23: what is/are the implications of this flag in terms of latency?

vectorstores/mongovector/mongovector_test.go
```diff
@@ -0,0 +1,592 @@
...
+	testIndexDP3              = "vector_index_dotProduct_3"
+	testIndexSize1536         = 1536
+	testIndexSize3            = 3
+)
+
+func TestMain(m *testing.M) {
+	flag.Parse()
+
+	defer func() {
+		os.Exit(m.Run())
```
tmc, line 42: this is pretty odd IMO, can we skip this defer and just wrap the setup in a conditional vs the early return?

tmc (commented):
Generally LGTM! have some small nits/asks.

Can we get an example (or two? or three?) together to showcase this?

tmc:
we should also get 'go test' to just work via a testcontainer container

tmc:
> we should also get 'go test' to just work via a testcontainer container

Oh I didn't realize this was leveraging a feature not in the F/OSS version of mongo -- is there a reasonable way to fake this so we don't take on a network dependency in CI?

PR #1011, opened 2024-09-12
fixes #1010 (anthropic not handling multipart tool responses)

### LINKED ISSUE #1010: anthropic driver doesn't handle multi-part tool results
Specifically [on this line](https://github.com/tmc/langchaingo) it only encodes the first part. The LLM is smart enough to make a second request, but it's wasteful.

lonelycode:
fixes #1010 

### PR Checklist

### DIFF
--- a/llms/anthropic/anthropicllm.go
+++ b/llms/anthropic/anthropicllm.go
@@ -272,35 +272,52 @@ func handleHumanMessage(msg llms.MessageContent) (anthropicclient.ChatMessage, e
 	return anthropicclient.ChatMessage{}, fmt.Errorf("anthropic: %w for human message", ErrInvalidContentType)
 }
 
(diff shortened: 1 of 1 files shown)


tmc (commented):
can you share a test or example showing an issue. 

I'm open to getting in a fix here but we really need more robust testing here.

bloeys:
Just wanted to note that while this (and related commits by @lonelycode) do solve the problem and give a proper tool response, I notice tool calls are _not_ streamed. This can be extremely helpful when tool arguments are very long.

We can either include a new 'WithToolStreamingFunc' or send tool chunks into the existing streaming function with perhaps some prefix (risk of false-positive), or just update the existing function signature (breaking change, but should be fine as we are still v0.x)

PR #1023, opened 2024-09-15
agents: fix the function call of openai llm.

chainhelen:
According to the [guides](https://platform.openai.com/docs/guides/function-calling) of openai function call, `Once you've executed these function calls in your application, you can provide the result back to the model by adding one new message to the conversation for each function call, each containing the result of one function call, with a tool_call_id referencing the id from tool_calls`. We need to provide the function call result back to the model in line with the specified structure, otherwise the openai server will verify the parameters and return a 400 status code.

### PR Checklist

### DIFF
--- a/agents/executor_test.go
+++ b/agents/executor_test.go
@@ -2,7 +2,9 @@ package agents_test
 
 import (
 	"context"
+	"fmt"
 	"os"
+	"slices"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


chainhelen:
If I use agents with function call by openai llm gpt-4o-2024-05-13, will get 400 status code, and the TestExecutorWithOpenAIFunctionAgent also cause the error
image.
This commit fix this

avi3tal:
+1 for this fix. otherwise openai agent is broken

PR #1027, opened 2024-09-21
googleapi/vertex: fix multi-tool call error, Invalid argument

CloudAceEmma:
### PR Checklist

This PR fix Invalid argument error when using multiple tools,  google genai only support one tool, multiple tools must be embedded into function declarations, see also:
https://github.com/GoogleCloudPlatform/generative-ai/issues/636
https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/function-calling#chat-samples


llms/googleai/googleai.go
```diff
@@ -373,7 +373,8 @@ DoStream:
 // convertTools converts from a list of langchaingo tools to a list of genai
 // tools.
 func convertTools(tools []llms.Tool) ([]*genai.Tool, error) {
-	genaiTools := make([]*genai.Tool, 0, len(tools))
+	genaiTools := make([]*genai.Tool, 0, 1)
```
eliben, line 376: This code now just creates a single-element slice `genaiTools` -- can it be simpler? You don't need to `make` the slice and then `append` a single element to it on the next line

(the same comment applies to the other file where this is done)

llms/googleai/shared_test/shared_test.go
```diff
@@ -373,6 +374,23 @@ func testTools(t *testing.T, llm llms.Model) {
 				},
 			},
 		},
+		{
+			Type: "function",
```
eliben, line 378: It would be very nice if the two functions/tools would be different - otherwise these two are identical and how will the model distinguish between their functionalities?

eliben (changes requested):
Thank you for the contribution!

initial comments

llms/googleai/googleai.go
```diff
@@ -450,9 +450,10 @@ func convertTools(tools []llms.Tool) ([]*genai.Tool, error) {
 		}
 		genaiFuncDecl.Parameters = schema
 
-		genaiTools = append(genaiTools, &genai.Tool{
-			FunctionDeclarations: []*genai.FunctionDeclaration{genaiFuncDecl},
-		})
+		// google genai only support one tool, multiple tools must be embedded into function declarations:
+		// https://github.com/GoogleCloudPlatform/generative-ai/issues/636
+		// https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/function-calling#chat-samples
+		genaiTools[0].FunctionDeclarations = append(genaiTools[0].FunctionDeclarations, genaiFuncDecl)
```
eliben, line 456: Perhaps I'm missing something obvious, but why do you need to predeclare `genatTools` and then do this `append`, can you not just create a single slice literal initialized to this?

CloudAceEmma: > Perhaps I'm missing something obvious, but why do you need to predeclare `genatTools` and then do this `append`, can you not just create a single slice literal initialized to this?

I just reused the `genaiTools` declaration before the change, now I initialized `genaiFuncDecls` and append all the funcs, `genaiTools` now initializes at the end.

eliben (approved):

eliben:
LGTM

@tmc to merge

nilpntr:
When will this be merged, struggling with the same issue?

nilpntr:
Any updates @CloudAceEmma @eliben @tmc ?

PR #1036, opened 2024-09-24
ollama: use official client

treywelsh:
### PR Checklist


pdxrlj:
I look forward to it, and I can use it together, so when can I use it?

treywelsh:
There's a problem with this PR, the `llms.CallOptions` doesn't allow to differentiate the cases between when an option is set to 0 or not set (for numeric types int, float64...).
The official ollama client take options via the input type `map[string]any`, so when building the map (via `makeOllamaOptionsFromOptions`) I finally have a temperature, penalties etc. with a value of 0 even if the user didn't set the option.
The langchaingo "internal" clients seems to Marshal a struct with some `omitempty` markers at some point, which remove empty values, or copy the empty value.

Infortunately, refactoring the generic llms package would imply to refact all clients options management.

PR #1044, opened 2024-10-04
tools: add pubmed tool

### LINKED ISSUE #1043: tools: add PubMed
Add PubMed as a tool like LangChain to make RAG more powerful.

longkeyy:
### PR Checklist

- [Y] Read the [Contributing documentation](https://github.com/tmc/langchaingo/blob/main/CONTRIBUTING.md).
- [Y] Read the [Code of conduct documentation](https://github.com/tmc/langchaingo/blob/main/CODE_OF_CONDUCT.md).
- [Y] Name your Pull Request title clearly, concisely, and prefixed with the name of the primarily affected package you changed according to [Good commit messages](https://go.dev/doc/contribute#commit_messages) (such as `memory: add interfaces for X, Y` or `util: add whizzbang helpers`).
- [Y] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [Y] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `Fixes #123`).
- [Y] Describes the source of new concepts.
- [Y] References existing implementations as appropriate.
- [Y] Contains test coverage for new functions.
- [Y] Passes all [`golangci-lint`](https://golangci-lint.run/) checks.

Fixes #1043


tools/pubmed/pubmed.go
```diff
@@ -0,0 +1,66 @@
...
+	"context"
+	"errors"
+
+	"github.com/tmc/langchaingo/callbacks"
+	"github.com/tmc/langchaingo/tools"
+	"github.com/tmc/langchaingo/tools/pubmed/internal"
+)
+
+// DefaultUserAgent defines a default value for user-agent header.
+const DefaultUserAgent = "github.com/tmc/langchaingo/tools/pubmed"
```
FluffyKebab, line 13: It think it would be better if this isn't exported and was set by the new function if the userAgent string is empty.

tools/pubmed/pubmed.go
```diff
@@ -0,0 +1,66 @@
...
+type Tool struct {
+	CallbacksHandler callbacks.Handler
+	client           *internal.Client
+}
+
+var _ tools.Tool = Tool{}
+
+// New initializes a new PubMed Search tool with arguments for setting a
+// max results per search query and a value for the user agent header.
+func New(maxResults int, userAgent string) (*Tool, error) {
```
FluffyKebab, line 25: What do you think about having functions options here instead? Then we could have a default for maxResults and the ability to pass callbacks.Handler in the New function.

tools/pubmed/pubmed_test.go
```diff
@@ -0,0 +1,159 @@
...
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/tmc/langchaingo/tools/pubmed/internal"
+)
+
+func TestNew(t *testing.T) {
```
FluffyKebab, line 16: Does these test need to be mocked? It seems like the tests test the internal client. Maybe we can put this file in the internal package.

PR #1059, opened 2024-11-04
anthropic: Support anthropic caching

lowjiansheng:
### PR Checklist


llms/options.go
```diff
@@ -272,11 +275,3 @@ func WithMetadata(metadata map[string]interface{}) CallOption {
 		o.Metadata = metadata
 	}
 }
-
-// WithResponseMIMEType will add an option to set the ResponseMIMEType
-// Currently only supported by googleai llms.
-func WithResponseMIMEType(responseMIMEType string) CallOption {
```
tmc, line 278: why pull this, this would be a breaking change.

llms/generatecontent.go
```diff
@@ -17,7 +17,10 @@ type MessageContent struct {
 }
 
 // TextPart creates TextContent from a given string.
-func TextPart(s string) TextContent {
```
tmc, line 20: We can't break the published api in this manner, this would need to be a variadic option list instead.

PR #1068, opened 2024-11-19
documentloaders: ms office docs

Struki84:
I've added a document loader that will read and parse MS office file types, .doc, .docx, .xls, .xlsx, .ppt, .pptx. Turns out it's a bit more complicated than I expected so I didn't extract the texts but all the doc data and parse it in the schema.Document.PageContent. 

For excel files there is metadata that will extract sheets and numerate them. The docx and pptx are just xml so I didn't extract the text just dumped the xml into PageContent, so at later date maybe somehow who understands the file formats better than me can build a decent document structure into schema.Document{}. 

At the same time I think there are some advantages of llm having access to entire document structure not just the text strings. 

 

### PR Checklist

- [ x] Read the [Contributing documentation](https://github.com/tmc/langchaingo/blob/main/CONTRIBUTING.md).
- [ x] Read the [Code of conduct documentation](https://github.com/tmc/langchaingo/blob/main/CODE_OF_CONDUCT.md).
- [x ] Name your Pull Request title clearly, concisely, and prefixed with the name of the primarily affected package you changed according to [Good commit messages](https://go.dev/doc/contribute#commit_messages) (such as `memory: add interfaces for X, Y` or `util: add whizzbang helpers`).
- [ x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [ x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `Fixes #123`).
- [x ] Contains test coverage for new functions.


documentloaders/ms_office.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+func (loader Office) loadDoc() ([]schema.Document, error) {
+	doc, err := mscfb.New(io.NewSectionReader(loader.reader, 0, loader.size))
+	if err != nil {
+		return nil, fmt.Errorf("failed to read DOC file: %w", err)
+	}
+
+	var text strings.Builder
+	for entry, err := doc.Next(); err == nil; entry, err = doc.Next() {
```
FluffyKebab, line 74: Can an error be returned by doc.Next() because of an other reason then there being no more entries?

documentloaders/ms_office.go
```diff
@@ -0,0 +1,228 @@
...
+			buf := make([]byte, entry.Size)
+			i, err := doc.Read(buf)
+			if err != nil {
+				return nil, fmt.Errorf("error reading WordDocument stream: %w", err)
+			}
+			if i > 0 {
+				// Process the binary content
+				for j := 0; j < i; j++ {
+					// Extract readable ASCII text
+					if buf[j] >= 32 && buf[j] <= 126 {
```
FluffyKebab, line 86: I think these numbers be unexported constants

Struki84: Got it!

documentloaders/ms_office.go
Struki84: Umm that's a good question. I'll have to check and get back to you with that. Will submit fix next week ;)

Struki84: doc.Next can return other types of error, added necessary checks

documentloaders/ms_office.go
Struki84: Done

rnarma321:
Hi, I was tinkering with this MR since I needed to use langchain Go to vectorize some .doc files and I found a "security vulnerability" my system was bugging me about.
It was complaining about the Doc.Read of line 128 leaves space for it flagging an error when it really is just an EOF not an error. I switched up the code to look like 
`func (loader Office) readEntryContent(doc *mscfb.Reader, entry *mscfb.File) (string, error) {
	buf := make([]byte, entry.Size)
	i, err := doc.Read(buf)

	if err != nil {
		if err == io.EOF {
			Log.Debug("input end reach")
		} else {
			return "", fmt.Errorf("error reading content stream: %w", err)
		}
	}
	if i > 0 {
		return extractTextFromBuffer(buf, i), nil
	}
	return "", nil
}`
I just added the if statement inside the error checking to validate it is not a io.EOF error.
Cheers

PR #1070, opened 2024-11-21
chains: Add llm multi prompt support+example

jubeless:
This PR introduces support for multi-prompt calls in llms, while maintaining backward compatibility with the LLMChain implementation for the call function. That said we believe the default of `LLMChain.Call` behaviour *should not* flatten prompts into a single message.

### Key Features
- Add GenerateFromMultiPrompt support in llms.
- Ensure backward compatibility for chains.Calls when using GenerateFromMultiPrompt.

### Backward Compatibility
No breaking changes have been introduced. Tracing is optional and does not affect existing functionality when disabled.

### PR Checklist

### DIFF
--- a/chains/llm.go
+++ b/chains/llm.go
@@ -24,7 +24,7 @@ type LLMChain struct {
 	CallbacksHandler callbacks.Handler
 	OutputParser     schema.OutputParser[any]
 	// When enabled usesMultiplePrompts will not 'flatten' the prompt into a single message.
-	UseMultiPrompt bool
+	useMultiPrompt bool
 
 	OutputKey string
 }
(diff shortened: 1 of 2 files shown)


chains/llm.go
```diff
@@ -11,14 +12,19 @@ import (
 	"github.com/tmc/langchaingo/schema"
 )
 
-const _llmChainDefaultOutputKey = "text"
+const (
+	_llmChainDefaultOutputKey     = "text"
+	_llmChainMultiPromptOutputKey = "choices"
```
FluffyKebab, line 17: Is it necessary to change the output key of the chain by default when using multipromt?

chains/llm.go
```diff
@@ -56,17 +68,27 @@ func (c LLMChain) Call(ctx context.Context, values map[string]any, options ...Ch
...
-	if err != nil {
-		return nil, err
+	llmsOptions := getLLMCallOptions(options...)
+	var llmOutput any
+	if c.useMultiPrompt {
+		llmsReponse, err := c.LLM.GenerateContent(ctx, chatMessagesToLLmMessageContent(promptValue.Messages()), llmsOptions...)
+		if err != nil {
+			return nil, fmt.Errorf("llm generate content: %w", err)
+		}
+
```
FluffyKebab, line 78: It should be documented somewhere that output parsers does not work with multiprompt.

tmc:
@jubeless could you address the PR feedback above?

PR #1125, opened 2025-02-02
feat: implement StreamingReasoningFunc for reasoning models

douglarek:
### PR Checklist

- [x] Read the [Contributing documentation](https://github.com/tmc/langchaingo/blob/main/CONTRIBUTING.md).
- [x] Read the [Code of conduct documentation](https://github.com/tmc/langchaingo/blob/main/CODE_OF_CONDUCT.md).
- [x] Name your Pull Request title clearly, concisely, and prefixed with the name of the primarily affected package you changed according to [Good commit messages](https://go.dev/doc/contribute#commit_messages) (such as `memory: add interfaces for X, Y` or `util: add whizzbang helpers`).
- [x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `Fixes #123`).
- [x] Describes the source of new concepts.
- [x] References existing implementations as appropriate.
- [x] Contains test coverage for new functions.
- [x] Passes all [`golangci-lint`](https://golangci-lint.run/) checks.

PR #1129, opened 2025-02-03
openai: support for "reasoning_effort" for o1/o3 models

### LINKED ISSUE #1122: Support `reasoning_effort` param for openai
New with the o3-mini model, right now we don't have control over this.

There should probably be some `WithCustomProperty(string, any)` so we don't get stuck behind when new properties are introduced and can at least immediately use them without waiting for a PR to make it type safe

Hades32:
Adds support of the reasoning effort field for reasoning models.

See https://platform.openai.com/docs/api-reference/chat/create#chat-create-reasoning_effort

Fixes #1122

### PR Checklist

### DIFF
--- a/llms/openai/internal/openaiclient/chat.go
+++ b/llms/openai/internal/openaiclient/chat.go
@@ -74,6 +74,9 @@ type ChatRequest struct {
 
 	// Metadata allows you to specify additional information that will be passed to the model.
 	Metadata map[string]any `json:"metadata,omitempty"`
+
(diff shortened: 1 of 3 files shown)


Hades32:
@tmc do you agree with the approach here, or should this be on the model level?

ppipada:
@Hades32 Note: I have opened a PR for supproting reasoning in anthropic and openai, which is based on some work of this PR and [1125](https://github.com/tmc/langchaingo/pull/1125)

PR #1135, opened 2025-02-06
fix: improve bedrock provider extraction from model id

lucaronca:
Currently, the bedrock implementation assumed that the model ID would start with the provider name, which meant it can't handle cases like passing the ARN as the model ID or where model ID uses a different format. For instance, AWS customer support recommends using a custom-formatted model ID to work around a bedrock bug (see: [AWS Repost](https://repost.aws/questions/QUEU82wbYVQk2oU4eNwyiong/bedrock-api-invocation-error-on-demand-throughput-isn-s-supported)). I have now updated the code to extract the provider using a more flexible approach that supports these custom formats.

### PR Checklist


karsharma10:
Would it be possible to get a review on this? I would like to see it in. I am facing the same issue with langchain go. Thank you!

sugkrish:
Thanks for sharing the workaround for inference profile ARNs. Would like to see this in as well. 😄

RicePatrick:
@tmc  - sorry for the ping, but is there an opportunity to get this reviewed/merged? It seems like a fairly safe change, and would greatly help those of us using bedrock 😄

lucaronca:
@tmc We also have this fix for bedrock which is quite important, let me know if it is ok. Thank you!

tmc (approved):
LGTM

tmc:
> @tmc We also have this fix for bedrock which is quite important, let me know if it is ok. Thank you!

Yep sorry about the delay, LGTM and merged.

cHiv0rz:
Do you know when is expected to generate a new version that contains this change?

RicePatrick:
@tmc  - any chance of a new version cut soon so that we would be able to use this? :pray:

PR #1141, opened 2025-02-09
anthropic: Add support for multi content part and images content in human messages.

Neofox:
As of now there wasn't an option to add an image to a Human message to the anthropic client.
Only text chat messages could be sent.

This PR will resolve two issues of the current implementation of `handleHumanMessage`.
It will also add the `anthropic-vision-example` example 

1. Allowing multi part content.

Looking at the [documentation](https://docs.anthropic.com/en/api/messages) we can see that there is two way of sending content through the API:

```
[
  {"role": "user", "content": "Hello there."},
]
```

or 
```
{"role": "user", "content": [
  { "type": "text", "text": "Hello there" },
  {"type": "text", "text": "How are you?"}
]}
```

but only the first case was currently supported.

2. Allowing one of the part to be an image

One of the content type that can be set is `image`
example:

```
{"role": "user", "content": [
  {
    "type": "image",
    "source": {
      "type": "base64",
      "media_type": "image/jpeg",
      "data": "/9j/4AAQSkZJRg...",
    }
  },
  {"type": "text", "text": "What is in this image?"}
]}
```

This feature will be also implemented with this PR, allowing to send Human Chat messages that also integrate images.
I actually needed this feature for my own project, this is how I discovered and fixed the issue.

### PR Checklist


tmc (approved):
Thanks for your contribution.

PR #1228, opened 2025-04-15
feat: Support Reasoning in anthropic and openai requests.

ppipada:
- Add Reasoning type that can be used by all LLMs to support reasoning. Add relevant flavors of reasoning types.
    - Support Hybrid budget token based reasoning in anthropic.
    - Support thinking content with and without streaming in anthropic.
    - Support level based reasoning request in openai
    - Need to make Temperature and TopP optional in internal structs in anthropic and openai.

### DIFF
--- a/llms/anthropic/anthropicllm.go
+++ b/llms/anthropic/anthropicllm.go
@@ -136,16 +136,48 @@ func generateMessagesContent(ctx context.Context, o *LLM, messages []llms.Messag
 	}
 
 	tools := toolsToTools(opts.Tools)
+	var thinking *anthropicclient.Thinking
+	temperature := &opts.Temperature
+	topP := &opts.TopP
+	// Retain current behaviour of omitting zero values from payload.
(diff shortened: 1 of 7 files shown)


ppipada:
@douglarek @tmc If possible, can you please have a look at this? There seems to be multiple PRs open related to reasoning support solving some subset of things. I have tried to solve this end to end. 

Thanks you for your work and time!

douglarek:
Thanks for the invitation. Although I haven't looked closely at each modification, I think destructive changes like those to fields such as CompletionRequest should be avoided as much as possible. After all, this is an exportable structure, and unnecessary disruptions to user space should be minimized.

ppipada:
@douglarek While I agree to the thought, but I am not sure how to achieve that if we want to support reasoning models in the openai and anthropic api's. The APIs actually error out (rather than ignore) if temperature value is present in the request. Any different workaround you see to solve for this?

PR #1231, opened 2025-04-22
Fix vertex ai multiple tool calls

nilpntr:
This PR addresses an issue in LangChain Go's `llms.WithTools()` function when used with Vertex AI LLMs.  Currently, when multiple tools are provided, the function incorrectly passes them to the Vertex AI API, resulting in an error.  Vertex AI expects multiple tools to be included within the function_declarations field of a single genai.Tool object.

This PR modifies the code to correctly format the tools for Vertex AI, ensuring that multiple tools are properly handled and the error is resolved. 

This is the current error message vertex provides: `rpc error: code = InvalidArgument desc = At most one tool is supported.`.

Link to vertex docs that tells us that at most one tool is supported if multiple put them all in a single genai.Tool object function_declarations: https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/function-calling#function-declarations


nilpntr:
Totally missed this one: https://github.com/tmc/langchaingo/pull/1027

PR #1244, opened 2025-05-13
googleai: fix issue where gemini would ignore all but the first tool

jpoz:
Fixes: https://github.com/tmc/langchaingo/issues/1245


tmc (approved):
We sorely need more comprehensive testing for exactly this. Thanks for the fix.

PR #1253, opened 2025-05-15
openai: add user and parallel tool calls fields to chat completions request

luiscamara:
### PR Checklist

---

The OpenAI LLM currently doesn't support two options:
- user
- parallel_tool_calls

The user field allows the consumer to send a unique identifier so OpenAI can track abuse.
The parallel_tool_calls allows the consumer to enable the execution of function calls in parallel when using a tool.

User seems to be very specific to OpenAI, so it makes sense to be in the LLM implementation and not a general setting.
Parallel tool calls are supported by more models, but after looking at the javascript langchain implementation, they also added a specific setting to the OpenAI LLM only.


universam1:
@tmc Please consider accepting this as we are also heavy dependent on the cost tracking support. Currently we see no other workaround here.

PR #1332, opened 2025-07-16
[BugFix] SSE comment line decoding error

rickif:
### PR Checklist
Fixes https://github.com/tmc/langchaingo/issues/942

https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events
>  Messages in the event stream are separated by a pair of newline characters. A colon as the first character of a line is in essence a comment, and is ignored.

The current implement returns error  `error decoding streaming response: invalid character ':' looking for beginning of value` when there're  SSE comments in the stream.
This PR ignores the SSE comments and avoid decoding error.

### DIFF
--- a/llms/openai/internal/openaiclient/chat.go
+++ b/llms/openai/internal/openaiclient/chat.go
@@ -447,11 +447,11 @@ func parseStreamingChatResponse(ctx context.Context, r *http.Response, payload *
 			}
 
 			// Ignore SSE comment line
-			if !strings.HasPrefix(line, ": ") {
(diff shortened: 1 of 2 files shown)


rickif:
https://openrouter.ai/docs/api-reference/streaming
> For SSE (Server-Sent Events) streams, OpenRouter occasionally sends comments to prevent connection timeouts. These comments look like: 
:OPENROUTER PROCESSING

If you're using OpenRouter, you've probably seen this a lot before.
@tmc Please take a look. It's quite important for the OpenRouter user. Thanks.

PR #1383, opened 2025-08-27
tools/scraper: fix context cancellation hanging

tmc:
Fixes an issue where the web scraper would hang indefinitely when the context is cancelled during the Wait() operation.

The scraper was checking for context cancellation before calling c.Wait(), but not during the wait operation itself. This caused the scraper to hang when a context was cancelled while scraping was in progress.

Solution:
- Moved c.Wait() into a separate goroutine
- Use select statement to handle both normal completion and context cancellation
- Ensures immediate return when context is cancelled

This change is complementary to PR #1295 which adds max pages limiting.


old-boys:
Any update for $TMC project, just obey the words what you have promised to the community,  have waited for couple months.

PR #1385, opened 2025-08-29
examples/groq: fixup groq example

tmc:
fixes/improves groq example by dropping dotenv and using a more interesting model


old-boys:
how about $TMC project, travis?

PR #1422, opened 2025-10-21
fix googleai recursive tool conversion: add missing fields in recursive tool conversion

sami-sweng:
### PR Checklist

Closes #1234, improves #1326 by adding missing parse of nullable and enum. Also better support what is thrown at it, for instance declaration parsed from yaml which are map[any]<smthg> by default.


layou233:
Can we have jsonschema support for Google AI? It just looks a subset of the genai.Schema, and it would help multiplatform development. 🥰

PR #1440, opened 2025-12-09
openaiclient: Fix `updateToolCalls` method for calls proxied with LiteLLM

dzbanek717:
When using LiteLLM as LLM proxy for accessing multiple models with one LLM agent in streaming mode the parameter `type: function` is always provided due to API conversion on LiteLLM side.

Despite the fact that it seems to be a problem on integration with LiteLLM side the whole problem lays in `updateToolCalls` function check. The assumption that `type` is not provided seems to be based more on the observation, rather than actual contract.

Basing on `function name` not being provided and part of `arguments` being provided this assumption seems to be also fine. Providing arguments for next tool shall always start with tool name. Therefore we are receiving parts of arguments until we received a new name. And it makes OpenAI Client compatible with both standard OpenAI API and LiteLLM proxy.

### PR Checklist

### DIFF
--- a/vectorstores/qdrant/rest.go
+++ b/vectorstores/qdrant/rest.go
@@ -19,7 +19,7 @@ func (s Store) upsertPoints(
 	ctx context.Context,
 	baseURL *url.URL,
 	vectors [][]float32,
-	payloads []map[string]any,
+	payloads []map[string]interface{},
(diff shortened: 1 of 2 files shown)


dzbanek717:
Hello @tmc , any chance on reviewing this?

PR #1480, opened 2026-02-25
llms/openai: openai client: return context.Canceled instead of new error declaration

mats852:
Returning the standard `context.Canceled` error from `llms/openai/internal/openaiclient/openaiclient.go` instead of a new declaration, so callers can evaluate it down the line.

Moreover, since cancelled can be spelled with both 1 and 2 `L`, it is error prone. Go's standard declaration uses 1 `L`.

before

```go
completion, err := llms.GenerateFromSinglePrompt(ctx, model, "prompt")

if err != nil && strings.Contains(err.Error(), "cancelled") {
	// handle canceled context
}
```

after

```go
completion, err := llms.GenerateFromSinglePrompt(ctx, model, "prompt")

if errors.Is(err, context.Canceled) {
	// handle canceled context
}
```

### PR Checklist

### DIFF
--- a/llms/openai/internal/openaiclient/openaiclient.go
+++ b/llms/openai/internal/openaiclient/openaiclient.go
@@ -225,7 +225,7 @@ func sanitizeHTTPError(err error) error {
 
 	// Check for context cancellation
 	if errors.Is(err, context.Canceled) {
-		return errors.New("request cancelled")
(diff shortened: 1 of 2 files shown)


mats852:
To prevent breaking implementation, we could join errors also, let me know what you think.

